PR #2, opened 2022-03-31
remove build-environment duplicates

sergiusens:
These are errors in snapcraft using core22.

Signed-off-by: Sergio Schvezov <sergio.schvezov@canonical.com>

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -55,7 +55,6 @@ parts:
     meson-parameters:
       - --prefix=/usr
       - --buildtype=release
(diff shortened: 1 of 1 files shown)


sergio-costas:
@sergiusens Is this patch still relevant? gnome-3-38-2004-sdk still has this bug, but this MR can't be merged. Meanwhile, gnome-42-2022-sdk doesn't have it...

sergiusens:
it should have been done against core22

sergio-costas:
Thanks.

Issue #6, opened 2022-04-29
Add libgtk3-nocsd.so.0 to gnome runtime Snaps for users who want to disable GTK3/4 CSDs


lucyllewy:
This library is combined with the `LD_PRELOAD` environment variable which gets set when installing the `gtk3-nocsd` deb package (this package also depends on the `libgtk3-nocsd0` package to install the actual `.so` library) on the host. Unfortunately the gnome runtime Snaps don't include the library so users are confused when a Snapped application uses CSDs despite the user having configured their system to disable them globally by installing the `gtk3-nocsd` deb package.

To implement this we should be able to "just" stage the `libgtk3-nocsd0` package in each runtime snap.

soumyaDghosh:
Hii,  @kenvandine  this issue seems to have been resolved long ago, shouldn't this be closed?

MorsMortium:
There is now libgtk-nocsd.so.0 (https://packages.ubuntu.com/search?keywords=libgtk-nocsd0) for all GTK versions

PR #8, opened 2022-04-29
Add libgtk3-nocsd0 to gnome-3-34-1804

lucyllewy:
Ref: #6
Signed-off-by: Dani Llewellyn <diddledani@ubuntu.com>


seb128:
Thank you for the contribution but the 3.34 snap isn't maintained anymore and we are not doing that kind of change to it now

PR #9, opened 2022-04-29
Add libgtk3-nocsd0 to gnome-3-38-2004

lucyllewy:
Ref: #6
Signed-off-by: Dani Llewellyn <diddledani@ubuntu.com>


3v1n0:
This is required by the ubuntu-unity and what else?

lucyllewy:
what else, any system where the user wants to disable CSDs. Not everyone likes them and we have the ability to turn them off on the host, but not in Snaps.

kenvandine (approved):
Thanks!

PR #12, opened 2022-05-18
Fix g-ir-introspection

sergio-costas:
When generating gobject-introspection, the python program g-ir-introspection is patched to find the python modules and other elements at $CRAFT_STAGE. Unfortunately, the environment name is inserted "as is" in the strings, instead of using os.environ to get the value.

This makes fail ATK, because when it tries to execute g-ir-scanner, it can't find one of the python modules that it needs to run.

This patch fixes this, by using os.environ.get('CRAFT_STAGE') instead.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -130,11 +130,7 @@ parts:
     override-build: |
       set -eux
       craftctl default
-      SCANNER=$CRAFT_PART_INSTALL/usr/bin/g-ir-scanner
-      sed -i "s#'/usr/lib#os.environ.get\('CRAFT_STAGE'\) + '/usr/lib#g" $SCANNER
-      sed -i 's#"/usr/lib#os.environ.get\("CRAFT_STAGE"\) + "/usr/lib#g' $SCANNER
-      sed -i "s#'/usr/share#os.environ.get\('CRAFT_STAGE'\) + '/usr/share#g" $SCANNER
-      sed -i 's#"/usr/share#os.environ.get\("CRAFT_STAGE"\) + "/usr/share#g' $SCANNER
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -131,8 +131,10 @@ parts:
       set -eux
       craftctl default
       SCANNER=$CRAFT_PART_INSTALL/usr/bin/g-ir-scanner
-      sed -i 's#/usr/lib#$CRAFT_STAGE/usr/lib#g' $SCANNER
-      sed -i 's#/usr/share#$CRAFT_STAGE/usr/share#g' $SCANNER
+      sed -i "s#'/usr/lib#os.environ.get\('CRAFT_STAGE'\) + '/usr/lib#g" $SCANNER
+      sed -i 's#"/usr/lib#os.environ.get\("CRAFT_STAGE"\) + "/usr/lib#g' $SCANNER
+      sed -i "s#'/usr/share#os.environ.get\('CRAFT_STAGE'\) + '/usr/share#g" $SCANNER
+      sed -i 's#"/usr/share#os.environ.get\("CRAFT_STAGE"\) + "/usr/share#g' $SCANNER
```
3v1n0: Can probably use something like:

```suggestion
      sed -i "s#\(["']\)/usr/\([^"']*\)#os.environ.get\('CRAFT_STAGE'\) + \1/usr/\2#g" $SCANNER
```

May be wrong, but to get the idea...

sergio-costas: Mmm... clever... Now I know to whom ask about RegExps ;-)

But it could fail if there are mixed ' and ", and also could replace paths that we don't want to replace. So I think that it could be better to keep it separated in two sed's, one for /usr/lib and another for /usr/share, but using your trick to detect both '  and " in a single line.

I'll do the change and test it.

sergio-costas: It seems to require escaping the " and the ', but otherwise, it seems to work like a charm. Also, I realized that there is no problem if " and ' are mixed, so your line should work fine in all cases.

PR #13, opened 2022-05-18
Test option doesn't exist in json-glib

sergio-costas:
The option 'test' isn't available in the meson_options.txt file, so it must be removed from the part.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -337,7 +337,6 @@ parts:
       - --prefix=/usr
       - --buildtype=release
       - -Ddocs=false
-      - -Dtests=false
     build-environment: *buildenv
 
   libpsl:


3v1n0:
LGTM, why is this depending on #12? I don't see merge conflicts.

sergio-costas:
Oh, sorry... It means that without #12 this patch doesn't work, due to build dependencies (because to compile json-glib, you need other packages that won't compile without #12). How should I rephrase it?

3v1n0:
Oh, fine... No need to specify it in the commit message, we can merge it also before given that we're buggy anyways

sergio-costas:
Ok, I'll remove that then.

sergio-costas:
Done.

PR #14, opened 2022-05-18
Libpsl can't find python

sergio-costas:
The psl-make-dafsa uses '/usr/bin/env python' to find the python runtime, but that searches for python2.

The documentation specifies that it requires python2.7+, ( https://github.com/rockdaboot/libpsl#readme ) which seems to indicate that it also does work with python3, so this patch changes the previous shebang line to /usr/bin/env python3.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -353,6 +353,10 @@ parts:
     build-packages:
       - libidn2-0-dev
       - libunistring-dev
+    override-pull: |
+      set -eux
+      craftctl default
+      sed -i 's#!/usr/bin/env python#!/usr/bin/env python3#g' src/psl-make-dafsa
 
   libsoup2:
     after: [ libpsl, meson ]
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -356,7 +356,7 @@ parts:
     override-pull: |
       set -eux
       craftctl default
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -353,6 +353,10 @@ parts:
     build-packages:
       - libidn2-0-dev
       - libunistring-dev
+    override-pull: |
+      set -eux
+      craftctl default
+      sed -i 's#!/usr/bin/env python#!/usr/bin/env python3#g' src/psl-make-dafsa
```
3v1n0: I'd be more confident in using patches here, as those are more targetting files.

So just include a patch file and use `patch -p1 < $file` to apply it.

sergio-costas: Ok.

sergio-costas: Mmm... how can I apply a patch that is outside the LXC container, in the snapcraft.yaml folder? In the documentation it says to use $SNAPCRAFT_STAGE, but that variable isn't set. Also, CRAFT_STAGE is set, but obviously points to /root/stage, inside the LXC container...

kenvandine: I think for this type of change a sed is fine, we do this in many cases within our desktop snaps.

PR #17, opened 2022-05-18
Fix file overlap in Gtk 4

sergio-costas:
Gtk 4 tries to stage three files that conflicts with others installed by the 'wayland' part. These files are:

    * usr/bin/wayland-scanner
    * usr/lib/x86_64-linux-gnu/libwayland-client.so.0.20.0
    * usr/lib/x86_64-linux-gnu/libwayland-server.so.0.20.0

Since these files really belong to 'wayland', it makes sense to don't stage the ones from 'gtk4'.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -497,6 +497,10 @@ parts:
       - libxi-dev
       - libxkbfile-dev
       - libxml2-utils
+    stage:
+      - -usr/bin/wayland-scanner
+      - -usr/lib/*/libwayland-client.so.0.20.0
+      - -usr/lib/*/libwayland-server.so.0.20.0
 
   gtk-locales:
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -496,6 +496,10 @@ parts:
       - libxi-dev
       - libxkbfile-dev
       - libxml2-utils
+    stage:
+      - -usr/bin/wayland-scanner
+      - -usr/lib/x86_64-linux-gnu/libwayland-client.so.0.20.0
```
kenvandine: It would be better to do this without the arch specific path, replace x86_64-linux-gnu with "*"

sergio-costas: Changed and tested.

kenvandine (approved):

PR #27, opened 2022-05-20
Updated package versions

sergio-costas:
* Added extra check when replacing '#!/usr/bin/env python' with '#!/usr/bin/env python3'
* Updated URIs for CoGL and Clutter-GTK (have been moved into Archive)
* Removed trailing blank spaces

Problems:

* Maximum version for GLibmm is 2.66.4 because 2.68 and newer requires libsigc++ >= 3, which isn't available in Ubuntu 22.04
* Maximum version for Cairomm is 1.14.3 because 1.16.0 and newer requires libsigc++ >= 3, which isn't available in Ubuntu 22.04
* Maximum version for Pangomm is 2.46.2 because 2.48.0 and newer requires libsigc++ >= 3, which isn't available in Ubuntu 22.04
* Maximum version for ATKmm is 2.28.2 because 2.36.0 and newer requires glibmm-2.68 or newer
* Maximum version for Gtkmm is 3.24.5 because 3.24.6 insists on building the documentation even after being told to do not

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1128,6 +1128,9 @@ parts:
       XML2_CONFIG=usr/bin/xml2-config
       sed -i 's#/root/parts/debs/install#/snap/gnome-42-2204-sdk/current#g' $XML2_CONFIG
 
+      rm -f usr/lib/vala-current
+      rm -f usr/share/gettext-current
+      rm -f usr/lib/$CRAFT_ARCH_TRIPLET/gdk-pixbuf-current
       ln -s vala-0.56 usr/lib/vala-current
       ln -s gettext-0.19.8 usr/share/gettext-current
       ln -s gdk-pixbuf-2.0/2.10.0 usr/lib/$CRAFT_ARCH_TRIPLET/gdk-pixbuf-current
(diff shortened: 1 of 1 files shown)


sergio-costas:
Should I try to update to versions with a bigger major version? Specifically:

* Update GtkSourceView from 4.8 to 5.4
* at-spi2-core from 2.44.1 to 3.6.3
* Libgnome-games-support from 1.8.2 to 2.0.0

sergio-costas:
BTW: the following parts doesn't have an specific branch or tag, so they are pulling from master/main:

* libsecret
* libinput
* libcanberra
* gsound
* libtool
* meson

Should I add one?

kenvandine:
> Should I try to update to versions with a bigger major version? Specifically:
> 
>     * Update GtkSourceView from 4.8 to 5.4
> 
>     * at-spi2-core from 2.44.1 to 3.6.3
> 
>     * Libgnome-games-support from 1.8.2 to 2.0.0

Yes, now's a good time for that

kenvandine:
> BTW: the following parts doesn't have an specific branch or tag, so they are pulling from master/main:
> 
>     * libsecret
> 
>     * libinput
> 
>     * libcanberra
> 
>     * gsound
> 
>     * libtool
> 
>     * meson
> 
> 
> Should I add one?

Yeah,let's pin to releases or stable branches where we can

kenvandine (approved):

PR #29, opened 2022-05-24
Fix meson build

sergio-costas:
Last snapcraft version requires that "meson" dependency be called "meson-deps".


kenvandine (approved):
Great work!

Issue #10, opened 2022-05-03
Add libadwaita to gnome-42 sdk


3v1n0:
We may need it to be included as it's part of GNOME apps, also we should include this patch: https://salsa.debian.org/gnome-team/libadwaita/-/blob/ubuntu/master/debian/patches/ubuntu/style-manager-Support-Yaru-accent-colors.patch

PR #37, opened 2022-06-30
Add libadwaita to Gnome-42-2204

sergio-costas:
Fix https://github.com/ubuntu/gnome-sdk/issues/10


3v1n0:
We also need [this patch](https://salsa.debian.org/gnome-team/libadwaita/-/blob/ubuntu/master/debian/patches/ubuntu/style-manager-Support-Yaru-accent-colors.patch)  to be applied to work with yaru.

Let me know if it need a rebase

sergio-costas:
Mmm... maybe that's why my SNAPs lack some icons...

sergio-costas:
I'll add this patch.

sergio-costas:
Added the patch.

sergio-costas:
This should be ready to be merged...

PR #38, opened 2022-07-07
Bug 1778359 - Produce debug symbols for GNOME snap

lissyx:


seb128:
Thanks, it sounds fine to me on principle.

I was unsure why the optimize was set to 3 so in case other reviewers wonder, https://mesonbuild.com/Builtin-options.html

buildtype release means debug=false and optimization=3 so the change is basically retaining the same level which was used before

lissyx:
> Thanks, it sounds fine to me on principle.
> 
> I was unsure why the optimize was set to 3 so in case other reviewers wonder, https://mesonbuild.com/Builtin-options.html
> 
> buildtype release means debug=false and optimization=3 so the change is basically retaining the same level which was used before

Yes, I pushed that late tonight to share, there are still a few dependencies within that where I'd like to add those flags.

Maybe I should also comment that the objcopy/strip parameters are mostly stolen from `dh_strip`'s.

sergio-costas:
Just a question: is it really a good idea to set both debug=true and optimization=3? (assuming that optimization=3 sets -O3). I mean: the optimizations in O3 are quite aggressive, so the debugging would be not very good.

Also, other distros discourage building things with -O3 because the code is bigger, in several cases it is slower (I presume that it is because being bigger it produces much more L1 cache misses), and can even break some programs... (although those can be old texts...)

Or maybe I'm too much conservative... O:-)

seb128:
@sergio-costas those are valid questions but probably worth another report, the merge request here is just a proposal to strip the debug symbol but -O3 is a meson default an already in use today right?

sergio-costas:
@seb128 Yes, you are right.

kenvandine:
@sergio-costas can you please look at this?

sergio-costas (approved):
It looks good to me.

PR #42, opened 2022-07-12
DT-465 Include meson/ninja in the SDK

sergio-costas:
This patch builds and includes meson and ninja in the SDK, thus
saving the maintainers of snaps to having to manually include it
in the snapcraft.yaml file.


kenvandine:
With this so snaps using the extension just set the plugin to meson and it works?  Or is there a change needed to snapcraft too?

sergio-costas:
> With this so snaps using the extension just set the plugin to meson and it works? Or is there a change needed to snapcraft too?

No change to snapcraft, it simply works.

kenvandine (approved):
Nice improvement

PR #43, opened 2022-07-12
DT-466 Remove libtool warning of files moved

sergio-costas:
When building libraries that depend on harfbuzz and glibmm,
libtool shows a warning saying that the libraries have been moved.
This MR fixes it.


kenvandine (approved):
Thanks

PR #46, opened 2022-07-20
Fix libsoup3

sergio-costas:
When compiling something with libsoup3, libnghttp2 is needed.
But the SDK lacks it, so although the libsoup-3.0.pc and
libraries are inside, pkg-config will say that it isn't available.

This MR fixes it.


sergio-costas:
Also other PKGs were failing, so I added some extra packages. Now all pkg-config tests do pass.

kenvandine (approved):
Thanks

PR #47, opened 2022-07-20
Fix broken links

sergio-costas:
There are several broken links in the SDK, specifically in the
usr/lib/TRIPLET folder. These broken links are:

- libc_malloc_debug.so
- libnsl.so
- libmount.so
- libuuid.so
- libblkid.so

This MR fixes them, and is complemented with another for the
CONTENTS snap.

The libc_malloc_debug.so link now points to the root.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1049,6 +1049,7 @@ parts:
       - gettext
       - itstool
       - libappindicator3-dev
+      - libblkid1
       - libbrotli-dev
       - libcairo2-dev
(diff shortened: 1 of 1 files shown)


kenvandine:
Is linking to libc_malloc_debug.so at the root valid in a snap build using the sdk snap?

sergio-costas:
@kenvandine as long as CORE22 gives the "base filesystem", yes, because it contains that file.

sergio-costas:
imagen

PR #51, opened 2022-07-21
Reduce snap size

sergio-costas:
This MR removes some files, like meson and ninja (unneeded in
the extension, only in the SDK) and strips all the binaries.
This reduce the final size of the snap.


kenvandine (approved):
Great!

PR #52, opened 2022-07-22
Adapt XXXmm versions to ensure useful building

sergio-costas:
When building gtkmm 3.24, if a too much recent version of glibmm
is detected, the builder will not use it and will download a
suitable old version. The same happens with cairomm and pangomm.

The net result is that we are compiling and adding code that
won't be used or needed, and that only takes space.

This MR adjusts the versions of all these parts to the last
valid version. Also adds some extra comments to avoid to
repeat this mistake in the future.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -652,7 +652,6 @@ parts:
     build-environment: *buildenv
     build-packages:
       - doxygen
-      - graphviz
 
   pangomm:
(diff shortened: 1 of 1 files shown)


sergio-costas:
It also reduces even further the final size of the CONTENTS snap.

kenvandine (approved):

PR #53, opened 2022-07-22
Reverts link to outside the snap

sergio-costas:
The store refuses to include a snap with a soft link outside its
own snap, so this line must be reverted. The solution is to copy
the file (which is fairly small, about 50Kbytes) during snap
building.


kenvandine (changes requested):
That would copy from the build host, you should copy that from the core snap

/snap/core22/current/lib/x86_64-linux-gnu/libc_malloc_debug.so.0

kenvandine (approved):
Great, thanks

PR #55, opened 2022-07-26
Store the debug symbols

sergio-costas:
This patch stores in a separate file the debug symbols, like the
https://github.com/ubuntu/gnome-sdk/pull/38 patch does for
Gnome38.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -259,6 +259,9 @@ parts:
     plugin: nil
     build-snaps:
       - gtk-common-themes
+    build-packages:
+      - python3-pip
+      - zip
     override-prime: |
       set -eux
 
@@ -270,4 +273,11 @@ parts:
       rm -rf usr/share/man
 
       find . -type d -empty -delete
-      $CRAFT_PROJECT_DIR/strip_binaries.py $CRAFT_PRIME
+
+      DEBUG_BUILDID=usr/lib/debug/.build-id
+      mkdir -p $DEBUG_BUILDID
+
+      python3 -m pip install pyelftools craft-parts
+      $CRAFT_PROJECT_DIR/strip_binaries.py $CRAFT_PRIME $DEBUG_BUILDID
+      echo Symbols compressed
+      rm -fr $DEBUG_BUILDID
--- a/strip_binaries.py
+++ b/strip_binaries.py
@@ -3,25 +3,58 @@
 import os
 import sys
 import subprocess
+from elftools.elf.elffile import ELFFile
+from elftools.elf.sections import NoteSection
+from elftools.elf.enums import ENUM_DT_FLAGS_1
+from craft_parts.sources.git_source import GitSource
+import yaml
 
 basepath = sys.argv[1]
+debugroot = sys.argv[2]
 
 files = {}
 
 paths = [basepath]
 
-# def read_retval(command):
-#     pr = subprocess.run(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-#     retval = pr.returncode
-#     if retval != 0:
-#         data = None
-#     else:
-#         data = pr.stdout.decode('utf8')
-#     return retval, data
+def describe_e_type(elffile):
+    x = elffile.header['e_type']
+    if elffile is not None and x == 'ET_DYN':
+        try:
+            # Detect whether this is a normal SO or a PIE executable
+            dynamic = elffile.get_section_by_name('.dynamic')
+            for t in dynamic.iter_tags('DT_FLAGS_1'):
+                if t.entry.d_val & ENUM_DT_FLAGS_1['DF_1_PIE']:
+                    return 'ET_PIE' # executable
+        except:
+            pass
+    return x
+
+def read_elf_type(filepath):
+    with open(filepath, "rb") as binary:
+        # find the BuildID
+        buildId = None
+        try:
+            elffile = ELFFile(binary)
+            ftype = describe_e_type(elffile)
+            for sect in elffile.iter_sections():
+                if not isinstance(sect, NoteSection):
+                    continue
+                for note in sect.iter_notes():
+                    if  note['n_type'] == 'NT_GNU_BUILD_ID':
+                        buildId = note['n_desc']
+                        break
+        except:
+            return None, None
+    return ftype, buildId
+
 
 while len(paths) != 0:
     path = paths[0]
+    if (path[-1] == os.path.sep) and (len(path) > 1):
+        path = path[:-1]
     paths = paths[1:]
+    if path == debugroot:
+        continue
     for filename in os.listdir(path):
         fullpath = os.path.join(path, filename)
         if os.path.islink(fullpath):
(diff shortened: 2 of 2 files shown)


kenvandine:
Since this is very similar to the PR for gnome-3-38, it would be great to have @lissyx review this.  @seb128 can you please take a look as well?

lissyx:
Why was a different patch made ? If my PR is not good enough, I can fix it ...

sergio-costas:
Your PR is for Gnome-3-38 snap, while this one is for Gnome-42. They are different branches. Also, since I was already stripping there the binaries, I just implemented the symbol storing directly in the python.

lissyx:
> Your PR is for Gnome-3-38 snap, while this one is for Gnome-42. They are different branches. Also, since I was already stripping there the binaries, I just implemented the symbol storing directly in the python.

Well, given firefox depends on gnome-3-38, I made the PR against that version because this is where we need it the most urgently.

Then we shoud just drop my PR and cherry-pick yours on gnome-3-38-sdk branch?

lissyx:
Ok, previous branches did not have the `strip_binaries`. It's also unclear to me why there are all those `-sdk` and not `-sdk` branches. This one

sergio-costas:
I don't think it is possible to do that: the gnome-3-38 branch is completely different and independent of the Gnome-42. @kenvandine...?

lissyx:
But `gnome-40-2004` branch does not seem to build any of the libs, I'm assuming it is build on the `-sdk` part? And this will be done on https://github.com/ubuntu/gnome-sdk/pull/54 ?

Does it means my current PR should be split the same way then ? Changing the build parameters to generate debug symbols on `gnome-3-38-2004-sdk` and collecting / strippping on `gnome-3-38-2004` ?

snapcraft.yaml
```diff
@@ -215,4 +219,19 @@ parts:
...
+      _VERSION=$SNAPCRAFT_PROJECT_VERSION
+      if [ "${_VERSION}" = "git" ]; then
+        pip3 install craft-parts
+        _VERSION=$(python3 -c "from craft_parts.sources.git_source import GitSource; print(GitSource.generate_version(part_src_dir='$CRAFT_PROJECT_DIR'))")
+      fi
+      SYMBOLS_ARCHIVE=$CRAFT_PROJECT_NAME_${_VERSION}_$SNAP_ARCH.debug
+      echo Compressing symbols
+      zip -r9 $CRAFT_PROJECT_DIR/$SYMBOLS_ARCHIVE $DEBUG_BUILDID
+      echo Symbols compressed
+      rm -fr $DEBUG_BUILDID
```
lissyx, line 283: This part seems mostly copy/paste of what I did, even though I am not proud of this, it serves its purpose

strip_binaries.py
```diff
@@ -5,23 +5,50 @@
...
-#     pr = subprocess.run(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-#     retval = pr.returncode
-#     if retval != 0:
-#         data = None
-#     else:
-#         data = pr.stdout.decode('utf8')
-#     return retval, data
+def read_retval(command):
+    # command must be a list of strings
+    pr = subprocess.run(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
lissyx: if you're find with a `subprocess` call, then it all looks just translation of the nasty bash I made into python, so it should be all fine

strip_binaries.py
```diff
@@ -5,23 +5,50 @@
...
+    retval, data = read_retval(['file', filepath])
+    if retval != 0:
+        return None, None
+    filetype = None
+    pos = data.find('BuildID')
+    if pos == -1:
+        BuildID = None
+    else:
+        BuildID = data[1+data.find('=',pos):data.find(',', pos)]
+    if ('LSB pie executable' in data) or ('LSB executable' in data):
```
lissyx: You might want to ensure `LC_ALL=C` is set, otherwise you're going to run into troubles (I did)

sergio-costas:
The Gnome-XX branches contain the snaps with the Gnome runtime for each Gnome version. They are needed for running programs. The Gnome-XX-SDK are for compiling programs and snapping them: they contain the compiler, libraries, headers and so on of an specific Gnome version. The Gnome-XX branches generate an snap with the same libraries than Gnome-XX-SDK, but without all the stuff needed to compile. This way, the programs don't need to include all the common Gnome libraries (like Gtk), but only those that are specific for them, or less used. And, of course, if you compiled a program with Gnome-3-38-SDK, it needs Gnome-3-38 snap; it can't run with, let's say, Gnome-42 snap.

Last week, trying to reduce the size of the Gnome-42 snap, we added binary stripping. And then we found your patch for Gnome-3-38 that stored the debug symbols in a separate file, and decided that the same should be done in Gnome-42 just in case they are needed. And that's why we modified the original patch to include storing the debug symbols also in Gnome-42.

I'm not sure if this answers your question...

lissyx (approved):

strip_binaries.py
sergio-costas: Yes, that's basically what I did.

lissyx:
> The Gnome-XX branches contain the snaps with the Gnome runtime for each Gnome version. They are needed for running programs. The Gnome-XX-SDK are for compiling programs and snapping them: they contain the compiler, libraries, headers and so on of an specific Gnome version. The Gnome-XX branches generate an snap with the same libraries than Gnome-XX-SDK, but without all the stuff needed to compile. This way, the programs don't need to include all the common Gnome libraries (like Gtk), but only those that are specific for them, or less used.
> 
> Last week, trying to reduce the size of the Gnome-42 snap, we added binary stripping. And then we found your patch for Gnome-3-38 that stored the debug symbols in a separate file, and decided that the same should be done in Gnome-42 just in case they are needed. And that's why we modified the original patch to include storing the debug symbols also in Gnome-42.
> 
> I'm not sure if this answers your question...

it does, I was waiting for a r+ on that before asking about other versions, to be honest, since I was first targetting what firefox uses, but it's clear having them on newer version will be useful either way, in the future for firefox or for others.

strip_binaries.py
lissyx: which is totally fine :)

sergio-costas: I was considering using an ELF module, but I preferred to be safe. But if you think that it would be better, I can do the change...

strip_binaries.py
sergio-costas: Good point!

strip_binaries.py
lissyx: It might, but honestly is it worth right now ? I'm unsure.

sergio-costas: Well... depending on the output of FILE can be dangerous (as you already said, it is a must to set "LC_ALL=C" to avoid problems), so maybe it is better...

lissyx: I wont block on it at least. I was wondering if a clean design would not be a snapcraft-level plugin, but I dont know anything about all of that and maybe it is a dumb idea.

sergio-costas: Ok, I used pyelftool.

strip_binaries.py
sergio-costas: Now I do everything in python, so this should not be a problem anymore.

strip_binaries.py
sergio-costas: @lissyx Can you review it and tell me what do you think?

lissyx: @sergio-costas I already r+'d it, and I had another look after seeing your push of `pyelftool` changes and I dont see anything that bothers me

sergio-costas: Thanks!

kenvandine (approved):
Thanks

PR #60, opened 2022-07-28
Fix python runtime

sergio-costas:
The python 3.10 runtime lacks several important files, so
executing it isn't possible unless using part of the resources
of core22. Unfortunately, since meson depends on python, not
having python means no meson too.

This patch fixes this by adding the basic package with the
modules.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1104,6 +1104,7 @@ parts:
       - libpython3-dev
       - libpython3.10-dev
       - libpython3.10-minimal
+      - libpython3.10-stdlib
       - libseccomp-dev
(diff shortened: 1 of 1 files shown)


sergio-costas:
Ok, I checked all the files in /usr/lib/python3.10, and all belong to these packages:

libpython3.10-stdlib*
libpython3.10-minimal*
python3-tk
python3-gdbm
libpython3.10-dev*
python3.10-venv
python3-lib2to3
python3-distutils

The three marked with an * are already in the yaml. I think that the others aren't needed, so this should be enough.

kenvandine (approved):

PR #63, opened 2022-08-02
DT-488 Add support for AMDGPU_ASIC_ID_TABLE_PATH

sergio-costas:
Every time a GTK-based program is launched in a computer with
an AMD-based GPU, the error

    /usr/share/libdrm/amdgpu.ids: No such file or directory

is shown in the terminal. The reason is that libdrm expects
that file to be in that precise location. Although the file
is now available in Gnome-42-2204, it isn’t accessible
directly, but each snap has to manually add a LAYOUT statement
to link that file from
$SNAP/gnome-contents/usr/share/libdrm to the root.

A possible fix would be to set that LAYOUT statement in
snapcraft, in the "gnome extension" code, but that means that
any snap that wants to include its own libdrm version won't be
able to put its own version of the file.

This patch fixes this by adding a new environment variable to
libdrm: AMDGPU_ASIC_ID_TABLE_PATH. This variable holds the
full path to the file, and it is used only when the file isn't
found in the expected location. Thus any snap that wants to
include its own version of libdrm can also use a LAYOUT
statement to place its own file at /usr/share/libdrm, while the
Gnome Extension snap can use the environment variable to point
to its file.

### DIFF
--- a/patches/libdrm.diff
+++ b/patches/libdrm.diff
@@ -1,0 +1,26 @@
+diff --git a/amdgpu/amdgpu_asic_id.c b/amdgpu/amdgpu_asic_id.c
+index a5007ffc..42b8f94c 100644
+--- a/amdgpu/amdgpu_asic_id.c
++++ b/amdgpu/amdgpu_asic_id.c
+@@ -112,12 +112,18 @@ void amdgpu_parse_asic_ids(struct amdgpu_device *dev)
+ 	ssize_t n;
+ 	int line_num = 1;
+ 	int r = 0;
++	char *amdgpu_asic_id_table_path;
+ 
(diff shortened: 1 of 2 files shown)


sergio-costas:
This patch has the libdrm version 2.4.110, which is the same that is available in Ubuntu 22.04.

sergio-costas:
I also uploaded the patch to the libdrm gitlab, but I'll wait to convert it into a MR until this MR has been discussed.

kenvandine:
Looks like libdrm.diff wasn't added

sergio-costas:
Added. Sorry.

sergio-costas:
Since, as commented in mattermost, it isn't supported to override libraries from the extensions, this patch has no meaning anymore. I'll do a new patch that just adds an entry for the layout.

Issue #49, opened 2022-07-21
More fonts are needed to avoid breaking rendering


lissyx:
As documented on https://bugzilla.mozilla.org/show_bug.cgi?id=1767925 the fontconfig is missing some bits. This results in Firefox Snap package showing poor rendering both of webpages as well as PDF documents.

Copying everything using the following ensures the same rendering on both sides:
```
mkdir ~/snap/firefox/current/.config/fontconfig/conf.d; cp /etc/fonts/conf.d/* ~/snap/firefox/current/.config/fontconfig/conf.d
```

![Screenshot_ubuntu22 04_2022-07-05_16:08:03](https://user-images.githubusercontent.com/1645737/180235974-1b7e5f98-c1a6-48a5-8f6a-2c186238285e.png)

lissyx:
![Screenshot_ubuntu22 04_2022-07-21_16:17:17](https://user-images.githubusercontent.com/1645737/180236294-e71b0621-038d-407a-8c82-444f5f1d3a4a.png)

seb128:
so the issue there is that some of the /etc/fonts/conf.d files are not available in the snap, the directory has some entries which symlinks to ../conf.avail, which are there in the snap but some other are pointing to /usr, example

/etc/fonts/conf.d/61-urw-nimbus-roman.conf -> /usr/share/fontconfig/conf.avail/urw-nimbus-roman.conf

could snapd perhaps make /usr/share/fontconfig/conf.avail/ also available in the snap env?

seb128:
I've also reported a snapd issue about it on https://bugs.launchpad.net/snapd/+bug/1982577

lissyx:
> 

So what would be a legit fix, modifying the `common/desktop-exports` script to take into account more fonts, or fix `snapd` itself ?

seb128:
> > 
> 
> So what would be a legit fix, modifying the `common/desktop-exports` script to take into account more fonts, or fix `snapd` itself ?

Since snapd already gives access to the conf.d/conf.avail it would probably make sense to include now the new directory in /usr/share/fontconfig, the snapd team seemed ok on principle with a such change.

jhenstridge:
We've avoided using the host system fontconfig configuration because it was prone to triggering incompatibilities. In particular, the config file parser treats any unknown element as a hard error.

[...]

PR #72, opened 2022-09-01
Include the fonts configuration from a default desktop installation

seb128:
We need at least the URW fonts configuration for correct PDF rendering as
reported against firefox on https://github.com/ubuntu/gnome-sdk/issues/49

After discussion we decided to include the configurations provided in a
default Ubuntu desktop installation for 3.38. It's not ideal but should
help with consistency with non snaps apps and be an improvement.
Going forward we should work with fontconfig upstream on a solution
which allows use to import the system configuration instead.


seb128:
I've tested the snap and the pdf example from #49 renders correctly now

kenvandine (approved):
Looks good

PR #74, opened 2022-09-20
Updated versions of libadwaita, glib, gtk4, and gtksourceview. Fixes #73

### LINKED ISSUE #73: Update libadwaita to 1.2
GNOME 43.0 is being released this week. Nearly every core GNOME app has switched to the new About dialog provided by libadwaita 1.2.

Instead of including a libadwaita part in all of our GNOME snaps, could we just update the version of libadwaita we provide in the gnome42 content snap to 1.2? It looks like it's buildable with Jammy's GTK4 4.6.

I think that's the only library update we need for the GNOME stuff we snap for GNOME 43.

Like usual, Epiphany will need a new webkitgtk but that should be coming via jammy-security before too long.

kenvandine:


sergio-costas:
There are several other updates:

ninja: v1.11.0 -> v1.11.1
meson: 0.62.2 -> 0.63.2
vala: 0.56.2 -> 0.56.3
gee: 0.20.5 -> 0.20.6
pango: 1.50.9 -> 1.50.10
libsoup3: 3.0.7 -> 3.0.8
poppler: 22.07.0 -> 22.09.0
libhandy: 1.6.0 -> 1.8.0
libgweather: 4.1.0 -> 4.1.1/4.2.0

Also, there are other possible updates, but I want to check compatibility:

libpeas: libpeas-1.32.0 -> libpeas-1.34.0
librsvg: 2.54.5 -> 2.55.1
harfbuzz: 4.4.1 -> 5.2.0

About updating GLib... I will ensure that it is safe, but in that case, I think that we should also update

gobject-introspection: 1.72.0 -> 1.74.0

jbicha:
@3v1n0 Is there any reason we wouldn't want glib 2.74 and gobject-introspection 1.74 in the gnome42 content snap?

sergio-costas:
BTW: remember that I wrote a tool to search for updates in SNAPCRAFT.YAML files: https://github.com/sergio-costas/updatesnap

sergio-costas:
Ok, librsvg can be updated to 2.55.1, because that's the new stable. They changed from odd->unstable, even->stable to the current Gnome way.

sergio-costas:
About harfbuzz, they say that they will never break the ABI, and after checking the deprecated symbols the most recently deprecated was in 3.4.0, so there should be no problem in updating it too.

sergio-costas:
Finally, the author of libpeas says that he doesn't want to break the ABI, so it should be safe to update libpeas too.

3v1n0:
> @3v1n0 Is there any reason we wouldn't want glib 2.74 and gobject-introspection 1.74 in the gnome42 content snap?

No, they're good to go from my POV.

---

Also, thanks for libadwaita updates

sergio-costas:
Today I did a new search, and there are these extra updates:

meson: 0.63.3
harfbuzz: 5.3.0
pango: 1.50.11
libsoup3: 3.2.0 and 3.1.4
wayland-protocols: 1.27
poppler: poppler-22.10.0
glibmm: 2.66.5
cairomm: 1.14.3
pangomm: 2.46.3
gtkmm: 3.24.7
gtksourceview: 5.6.1
GJS: 1.72.3

sergio-costas (approved):
LGTM

PR #75, opened 2022-09-20
Fixed missing deps found in lint

kenvandine:


sergio-costas:
Odd... shouldn't those be installed automagically due to dependencies?

Anyway, LGTM.

sergio-costas (approved):

kenvandine:
> Odd... shouldn't those be installed automagically due to dependencies?
> 
> Anyway, LGTM.

I would have expected so, but I guess not.

PR #76, opened 2022-09-22
Fix build

sergio-costas:
There seems to be a problem with the last version of craft-parts, so this patch fixes the version to the previous one.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1418,7 +1418,7 @@ parts:
       mkdir -p $DEBUG_BUILDID
 
       OLDPATH=$PATH
-      python3 -m pip install pyelftools craft-parts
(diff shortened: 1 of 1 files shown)


sergio-costas:
@kenvandine I tried to build gnome-42-sdk today, and found that it fails during the PRIME stage due to a problem installing the craft-parts python module. Can you check if it happens to you too? In that case, this patch seems to fix it.

sergio-costas:
It seems that there is a bug in the building of the last craft-parts version. I reported it: https://github.com/canonical/craft-parts/issues/283

kenvandine (approved):

sergio-costas:
When they fix the bug, we can revert this.

PR #78, opened 2022-10-03
Sdk indicator cleanup

kenvandine:


sergio-costas (dismissed):
LGTM

sergio-costas:
Mmm... wait... in my system, if I try to install libappindicator3-dev, it says that there is a conflict with libayatana-appindicator3. So I think that we should choose which one to install: or libappindicator3-dev, or libayatana-appindicator3-dev, but it seems that it is not possible to install both.

sergio-costas:
Ok, I built the sdk with this patch and it included both libraries, which means that programs linked against libappindicator will use it, and programs linked against libayatana-appindicator will use it, no matter that libayatana-appindicator is compatible with libappindicator.

Should we remove the libappindicator?

kenvandine:
Yeah, I think we should remove it

PR #1, opened 2022-03-31
Update README.md

sergiusens:

PR #81, opened 2022-10-21
Update README for Core22

sergio-costas:
This patch fully fixes the README.md for Gnome-42-2204-sdk.

It overrides https://github.com/ubuntu/gnome-sdk/pull/1


sergio-costas:
Ok, now I discovered how to do suggestions, so removing this.

PR #82, opened 2022-10-27
DT-777: apply check-portal-plugs-for-snap patch to Core20

sergio-costas:
This MR applies to Core20 the GLib MR https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3020

This MR makes GLib to use the keyfile backend with dconf also in Snaps, not only in Flatpak.


kenvandine (approved):
Looks good

sergio-costas:
@kenvandine Wow!!!! Really fast :-D

PR #83, opened 2022-10-27
DT-777: apply check-portal-plugs-for-snap patch to Core22

sergio-costas:
This MR applies to Core22 the GLib MR https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3020

This MR makes GLib to use the keyfile backend with dconf also in Snaps, not only in Flatpak


sergio-costas:
@kenvandine Ok, this MR has the sandbox patch incorporated into glib (and, of course, tested too). It also updates glib to the last version (1.74.3, which seems to fix the build problem with gjs).

kenvandine (approved):
Looks good, thanks

PR #88, opened 2022-11-24
Update to last versions

sergio-costas:
GLib can't be updated to 2.74.2 because it breaks GJS build, so it's updated only to 2.74.1

GJS can't be updated to 1.75.1 because it requires LibMozJS-102, but it isn't available in Jammy, so it's kept at the current version.

Poppler could be updated to 22.11, but in GIT, it didn't pass the tests.

LibSoup can be updated to 3.1.4 or to 3.2.2...

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -43,7 +43,7 @@ parts:
     after: [ ninja ]
     plugin: nil
     source: https://github.com/mesonbuild/meson.git
-    source-tag: '0.63.3'
+    source-tag: '0.64.1'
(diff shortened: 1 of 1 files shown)


sergio-costas:
About the problem with GLib, there are two options:

* either I can try to write a patch that fixes the problem,
* or build libmozjs-102 from sources to be able to use GJS (in the hope that it prevents this bug)

About Libsoup... Should we update to any of those versions?

kenvandine (approved):
LGTM, thanks

PR #89, opened 2022-12-02
Add tags for update_snap.py

sergio-costas:
This MR adds tags to allow update_snap.py to fully parse it automagically and generate a list with updates in an unnattended fashion.

update_snap repository: https://github.com/sergio-costas/updatesnap

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -76,6 +76,9 @@ parts:
     after: [ libtool, meson-deps ]
     source: https://gitlab.freedesktop.org/gstreamer/meson-ports/libffi.git
     source-tag: 'meson-3.2.9999.4'
+# ext:updatesnap
+#   version-format:
+#     format: 'meson-%M.%m.%R'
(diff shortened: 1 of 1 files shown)


sergio-costas:
Update_snap.py reads the same file than snapcraft; that means that it's not possible to add new tags without breaking the building process. But using a different file to specify things like the version format for each part, or other details like if odd-minor-version numbers are development versions and should be ignored, is cumbersome.

To solve this, the new data is added as comments, keeping the YAML format, and preceded by `# ext:updatesnap`. When `update_snap.py` loads the snapcraft.yaml file, it firsts scans it searching for those comment blocks, and whenever one is found, the first line with `# ext:updatesnap` is removed, and each ' #' symbol in the next lines are replaced by a space. This "recomposes" the YAML file and leaves it ready to be parsed by the standard YAML module, while preserving full compatibility with snapcraft.

hellsworth:
lgtm!

PR #91, opened 2022-12-03
Update metadata to point to source repo

merlijn-sebrechts:
I noticed is wasn't super clear where I could find the source for the content snaps, so I thought "why not add it to the metadata in the store?"

I also added some info to guide users and developers towards the right docs.


kenvandine (approved):
This is great, thanks

PR #93, opened 2022-12-07
Drop libappindicator3-1 as we already have libayatana-appindicator3-1

kenvandine:
Don't stage both, we use libayatana-appindicator3-1 now.  This probably fixes https://github.com/snapcrafters/mattermost-desktop/issues/66


sergio-costas (approved):
As already commented, it makes sense to remove the package.

PR #96, opened 2023-01-05
Build webp-pixbuf-loader

JGCarroll:
This patch allows GTK apps to be able to parse webp image files.

Overall this ships 2 new files with the snap:

./usr/share/thumbnailers/webp-pixbuf.thumbnailer
./usr/lib/x86_64-linux-gnu/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader-webp.so

The thumbnailer isn't necessarily required and it may be worth removing; however I've kept it as the content snap already seems to have two other thumbnailers, so it would be consistent with what's already there.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1251,6 +1251,19 @@ parts:
     build-packages:
       - gir1.2-glib-2.0
 
+  webp-pixbuf-loader:
+    after: [meson-deps, gdk-pixbuf]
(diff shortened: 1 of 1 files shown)


kenvandine:
Thanks for your contribution!  I think this looks good, but will ask for another review as well.

sergio-costas (approved):
It also looks good to me.

PR #97, opened 2023-01-11
Updated several parts

sergio-costas:
Some could also be updated to new minor versions:

- pixman, from 0.40.0 to 0.42.2
- libsoup3, from 3.0.8 to 3.2.2
- libpoppler, from 22.10.0 to 22.12.0
- libinput, from 1.21.0 to 1.22.0
- pycairo, from 1.21.0 to 1.23.0

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -104,7 +104,7 @@ parts:
   glib:
     after: [ libffi, meson-deps ]
     source: https://gitlab.gnome.org/GNOME/glib.git
-    source-tag: '2.74.4'
+    source-tag: '2.74.5'
 # ext:updatesnap
 #   version-format:
 #     ignore-odd-minor: true
@@ -134,7 +134,7 @@ parts:
   pixman:
     after: [ glib, meson-deps ]
     source: https://gitlab.freedesktop.org/pixman/pixman.git
-    source-tag: 'pixman-0.40.0'
+    source-tag: 'pixman-0.42.2'
 # ext:updatesnap
 #   version-format:
 #     format: 'pixman-%M.%m.%R'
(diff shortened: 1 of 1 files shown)


kenvandine:
@sergio-costas how risky do you think libsoup3 from 3.0.x to 3.2.x might be?  That's the only one there I might be hesitant on, but would be nice to update.

sergio-costas:
The documentation says "API version: 3.0, Library version: 3.3.0", so it seems that, at API level, it's compatible. But in the documentation they put an example of macros for both minimum and maximum version, so I'm not sure... I'll ask in the mailing list.

sergio-costas:
the mailing list server is down... I'll update the other in the meantime.

sergio-costas:
Ok, I uploaded another patch to this MR to update everything except libsoup (and harfbuzz, but that I want to do a test). Also, I asked about binary compatibility in Discourse.

sergio-costas:
Ok, tested. It works.

sergio-costas:
Ok, about libsoup, the answer is "It certainly should be", by Michael Catanzaro https://discourse.gnome.org/t/libsoup-binary-compatibility/13608

kenvandine (approved):
LGTM

kenvandine:
> Ok, about libsoup, the answer is "It certainly should be", by Michael Catanzaro https://discourse.gnome.org/t/libsoup-binary-compatibility/13608

Let's update libsoup3 in a separate PR

sergio-costas:
Perfect. About that, yesterday I found this page: https://abi-laboratory.pro/?view=compat_report&l=libsoup&v1=2.49.1&v2=2.50.0&obj=63366&kind=abi

It compares the ABIs between versions of a library to check if they are binary compatible. I think that it's a good idea, and it could be automatized for this. What do you think?

PR #98, opened 2023-01-20
Add -O3 and -g parameters to autotools parts

sergio-costas:
The parts built with the MESON module have been using the -O3 and -g parameters since several months ago, but the parts built with the AUTOTOOLS module weren't. This patch fixes this.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -73,7 +73,10 @@ parts:
     source: https://git.savannah.gnu.org/git/libtool.git
     source-tag: 'v2.4.7'
     plugin: autotools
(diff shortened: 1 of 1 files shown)


sergio-costas:
I'm not sure if for C++ code I have to use CCFLAGS or CXXFLAGS.

kenvandine:
> I'm not sure if for C++ code I have to use CCFLAGS or CXXFLAGS.

I think it's CXXFLAGS, but not certain.

sergio-costas:
Anyway, using CXXFLAGS, epiphany doesn't work, so it seems that using that is problematic. Closing.

PR #99, opened 2023-01-30
Update to last versions

sergio-costas:
This patch updates:

* poppler from 22.12.0 to 23.01.0
* libsoup3 from 3.0.8 to 3.2.2
* harfbuzz from 5.3.1 to 6.0.0

The three libraries have been tested with abi-compliance-checker to ensure that they are ABI compatible with the old versions.


kenvandine (approved):
✅

PR #100, opened 2023-01-31
Update harfbuzz and libpsl to meson

sergio-costas:
Both libraries were being built with autotools, but meson build was also available, which is preferred because is more modern and faster.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -304,28 +304,30 @@ parts:
     build-environment: *buildenv
 
   harfbuzz:
-    after: [ fribidi ]
+    after: [ fribidi, meson-deps ]
     source: https://github.com/harfbuzz/harfbuzz.git
     source-tag: '5.3.1' # developers declared that they won't break ABI
     source-depth: 1
-    plugin: autotools
-    autotools-configure-parameters:
+    plugin: meson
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -304,28 +304,30 @@ parts:
...
     source-depth: 1
-    plugin: autotools
-    autotools-configure-parameters:
+    plugin: meson
+    meson-parameters:
       - --prefix=/usr
-      - --with-graphite2=yes
-      - --enable-introspection
-      - --with-gobject
-      - --enable-static
```
kenvandine, line 317: I didn't realize we had enabled static before, I assume that's statically linked?  We probably didn't really want that, but with this change do you think we should be concerned we might break other builds?

sergio-costas: AFAIK, --enable-static doesn't link this statically, but just creates .a files to allow harfbuzz to be linked statically. I set "--default-library=both" in meson to build both dynamic and static libraries, so it will be possible to link it in both ways.

kenvandine: Excellent, thanks for the clarification

kenvandine (approved):
LGTM

PR #103, opened 2023-02-10
Add libdir to autotools builds

sergio-costas:
This ensures that the libraries are installed at /usr/lib/TRIPLET.

glibmm and atkmm aren't migrated, because they fail to build.


sergio-costas:
This still requires testing before merging.

sergio-costas:
@kenvandine It seems to work (I tested it with several VALA programs), but still want to do some extra tests to ensure that everything works as expected. Specifically I want to test snaps that use:

- librsvg
- librest
- libcanberra
- cogl
- gtkmm

Any suggestion for those?

PR #105, opened 2023-02-10
Add python3 executable

sergio-costas:
The original snap lacks the python3.10 executable, so it can't be used to run python programs. This patch adds
python3.10-minimal to include it.

It also reorders alphabetically some packages that were out of place.


sergio-costas:
It adds less than 3 megs to the final snap size.

kenvandine (approved):
Thanks!

PR #111, opened 2023-03-22
Use local gtk binaries

sergio-costas:
The "caches" part installs gtk-update-icon-cache, libglib2.0-bin and shared-mime-info .deb packages to use some of their binaries. Unfortunately, the $PATH puts before the built binaries instead of the ones in the system, but the libraries are from the system. This can result in a failure if a new version being compiled requires a new call in the libraries.

In fact, updating in gnome-sdk glib to the last version triggers this problem.

This patch fixes it.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -247,14 +247,16 @@ parts:
       - gtk-update-icon-cache
       - libglib2.0-bin
       - shared-mime-info
+    build-environment:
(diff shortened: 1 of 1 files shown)


kenvandine (commented):
Wouldn't it be better not to use the build-packages but only use the versions we are actually going to stage in the content snap?

sergio-costas:
Maybe setting LD_LIBRARY_PATH... I'll try it.

sergio-costas:
@kenvandine Done.

kenvandine (approved):
Looks good, thanks!

Issue #114, opened 2023-04-09
dependency issue for snaps built with gnome-42-2204-sdk candidate


jbicha:
Test Case
-------------
```sh
$ snap refresh --stable gnome-42-2204 # ensure that we are on the stable branch
$ snap install --candidate evince # version 1022
$ snap run evince
/snap/evince/1022/usr/bin/evince: symbol lookup error:
 /snap/evince/1022/usr/bin/evince: undefined symbol: g_string_free_and_steal
```

Suspected Issue
----------------------
- `g_string_free_and_steal` is a new symbol from glib 2.75
- gnome-42-2204-sdk was updated last month to include glib 2.76.0 (part of GNOME 44)
- Although evince does not directly use that symbol, when it was built against glib 2.76, it ended up with a dependency on it

Suggested Fixes
----------------------
- Revert glib back to 2.74.6 for now
  - And then rebuild evince candidate
- Or push the new gnome-42-2204 (and gnome-42-2204-sdk) to stable.
  - However, glib 2.76 is disruptive. One disruptive change was reverted in 2.76.1 (to be reintroduced for 2.77 early) so you probably want at least 2.76.1 if you are going ahead with 2.76.x
  - The other is the removal of the slice allocator. The Ubuntu Desktop distro team is considering reverting that change for the glib deb for Ubuntu 23.04 since it has triggered crashes in apps which have pre-existing serious bugs.

jbicha:
Thanks, the evince candidate snap is working now.

jbicha:
I'm going to close this issue now. If there were any other affected snaps, they just need to be rebuilt.

PR #116, opened 2023-04-10
Revert GLib to 2.74.6

sergio-costas:
Due to version 2.76.0 being a quite disruptive version, it seems a good idea to wait before updating it.


sergio-costas:
@kenvandine This MR reverts GLib and gobject-introspection to version 2.74. It also re-instates the patch for snap.

kenvandine:
This fixed #114

kenvandine (approved):
Looks good and probably wise to stick with 2.74 for a while

PR #117, opened 2023-04-10
Build libayatana-appindicator from source to get upstream fixes

kenvandine:
Upstream libayatana-appindicator has had some crash fixes which aren't backported for 22.04.  This builds it from source to get those fixes.


sergio-costas (approved):
Looks good. Also tested with epiphany, mattermost, discord and snap-store, and seems to work fine.

PR #118, opened 2023-04-10
Several updates.

sergio-costas:
libinput could be updated to 1.23.0, but preferred to wait.
meson could be updated to 1.1.0, but that version was launched today, so I think that it's better to wait.


sergio-costas:
@kenvandine Another update.

kenvandine (approved):

PR #119, opened 2023-04-10
Updated to work with libayatana-appindicator build in the sdk snap

kenvandine:


sergio-costas (approved):
I was afraid that removing "libayatana-appindicator3-1" and moving the stag snap to the stable one, the appindicator libraries would "disappear"; but no, they are there.

imagen

PR #124, opened 2023-04-21
SDK with GLib 2.76.1

sergio-costas:
This branch is for tests.


3v1n0:
FYI, I released yesterday 2.76.2, changes should be quite safe, so you can use that! :)

sergio-costas:
@3v1n0 The idea is to launch Gnome-42-2204 with GLib 2.76.x but keep Gnome-42-2204-SDK with 2.74.y; thus, if there are big compatibility problems, it would be safe to revert the base snap. But I'll update this right now. Thanks!

sergio-costas:
(I mean: to launch it that way for a while, to ensure that there are no big problems).

PR #128, opened 2023-05-04
Move scripts and other files into tools folder

sergio-costas:
To avoid plaguing the root folder with scripts and files, this MR moves the strip_binaries.py and the bindtextdomain.c files into the tools/ folder. The idea is to also put there any new script or file needed to build the snap.


kenvandine (approved):
Sensible change, thanks!

PR #129, opened 2023-05-08
DT-503 Compile Wacom library instead of using package

sergio-costas:


sergio-costas:
@kenvandine This MR was pending since, at least, October 2022, but for some reason I forgot to send it as a merge request. The task was in Jira, but not assigned to me. Sebastien found it, so I updated the repository to the last version and tested it, and it seems to work.

kenvandine (approved):
LGTM

PR #130, opened 2023-05-12
Updated parts to the last version

sergio-costas:
This MR updates the parts to the last version, and also adds the "source-depth" tag to all the parts to reduce the required bandwidth and disk space to build the snap.

Also, it adds some extra updatesnap tags.

It doesn't update GLib to version 2.76 yet.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -38,6 +38,7 @@ parts:
     plugin: nil
     source: https://github.com/ninja-build/ninja.git
     source-tag: 'v1.11.1'
+    source-depth: 1
     override-build: |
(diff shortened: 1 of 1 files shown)


soumyaDghosh:
Kindly update the gtk4 version to 4.11 because the latest gnome incubator apps like snapshot need that

sergio-costas:
> Kindly update the gtk4 version to 4.11 because the latest gnome incubator apps like snapshot need that

AFAIK, gtk 4.11 is a development version, so it should not be used in a stable system.

kenvandine (approved):
Thanks!

PR #131, opened 2023-05-15
Fix thumbnailer

sergio-costas:
This patch fixes the thumbnails in cheese.


kenvandine (approved):
👍

jbicha:
I think I rebuilt cheese against the updated gnome42-2204-sdk but I can't tell.

The thumbnails are not working in the cheese snap. Do you want me to open an issue somewhere for this? Where?

seb128:
you shouldn't need a rebuild of cheese but did we get a gnome-42-2204 build including the change?

kenvandine:
Not yet... But I'll kick that off

PR #134, opened 2023-05-18
DT-1290: Add libva, intel-gmm and intel-media-driver

sergio-costas:
This patch adds the libAV library with DRI3 support, the Intel Graphics Memory Manager library (needed for the driver) and the Intel Media Driver.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1427,8 +1427,62 @@ parts:
         sed -i 's#includedir=/usr#includedir=${prefix}#' $PC
       done
 
+  libva:
+    after: [ libtool, meson-deps, libayatana-appindicator ]
+    source: https://github.com/intel/libva.git
+    source-tag: '2.18.0'
+    source-depth: 1
+    plugin: meson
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -1427,8 +1427,62 @@ parts:
         sed -i 's#includedir=/usr#includedir=${prefix}#' $PC
       done
 
+  libva:
+    after: [ libtool, meson-deps, libayatana-appindicator ]
```
kenvandine, line 1431: Does libva really need to be after libayatana-appindicator?

kenvandine (changes requested):

snapcraft.yaml
sergio-costas: Not really... it was just to build it "in order".

sergio-costas:
Anyway, as I see, firefox still uses gnome-3-38...

kenvandine (approved):
Thanks, this is good to land

PR #135, opened 2023-05-18
Update to GLib 2.76

sergio-costas:
This PR updates GLib to version 2.76. It should be applied when it is considered safe.


sergio-costas:
@kenvandine If the "gnome-42-contents" snap with GLib 2.76 has been published and produced no negative reactions, then it should be safe to publish this...

kenvandine (approved):
I do think this is safe now

PR #137, opened 2023-05-19
Fix GLib build

sergio-costas:
When updating to GLib 2.76, the patch for snap must be removed, because it is already merged.


kenvandine (approved):
Thanks!

PR #140, opened 2023-05-30
Only build intel specific parts on amd64, fixes #139

### LINKED ISSUE #139: intel-gmm and intel-media-driver should only be biult for amd64
The intel-gmm and intel-media-driver parts are only useful for amd64 and actually fails to build on arm{hf,64} arches.  We should only build these parts for amd64.

kenvandine:
This should fix build failures on armhf and arm64


sergio-costas (approved):
lgtm.

PR #143, opened 2023-06-01
Include the fonts configuration from a default desktop installation

seb128:
We need at least the URW fonts configuration for correct PDF rendering as reported against firefox on https://github.com/ubuntu/gnome-sdk/issues/49

After discussion we decided to include the configurations provided in a default Ubuntu desktop installation for 42. It's not ideal but should help with consistency with non snaps apps and be an improvement. Going forward we should work with fontconfig upstream on a solution which allows use to import the system configuration instead.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -228,8 +228,16 @@ parts:
       find . -type f,l -exec rm -f $CRAFT_PART_INSTALL/usr/lib/{} \;
       find . -type f,l -name "*.so*" -exec bash -c "rm -f $CRAFT_PART_INSTALL/usr/lib/{}*" \;
 
+  fonts-config:
+      after: [ debs ]
+      plugin: nil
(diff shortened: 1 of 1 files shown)


seb128:
It's a cherrypick of !72 / 9290ff9 which we apparently didn't include in the new sdk version? 

It should fix #49 again for firefox/core22, the regression was reported as https://bugs.launchpad.net/ubuntu/+source/fontconfig/+bug/2017076. 

Note that the list of packages is slightly different since the fonts installed by default slightly changed between 20.04 and 22.04. I've also added language-selector-common to the list since Gunnar pointed out during the debug discussion that it also includes configuration. It seems logical but I don't remember now if we had it excluded for some reason? If we don't have any I would suggest we go with it and tweak later if problems get reported.

seb128:
One way to verify the change based from the firefox launchpad report is to do

`$ LC_CTYPE=zh_CN.UTF-8 fc-match`
in the snap shell environment

it should report
> NotoSansCJK-Regular.ttc: "Noto Sans CJK SC" "Regular"

kenvandine (approved):
Looks like a sensible change, thanks!

PR #145, opened 2023-06-04
Enable building gnome-42-2204-sdk on riscv64

xypron:
With the following changes the gnome-42-2204-sdk snap can be built on riscv64:

* run autogen.sh for libcanberra to replace the outdated config.guess
* use a newer version of libffi including a patch enabling riscv64 builds

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -942,6 +942,9 @@ parts:
       - --prefix=/usr
       - --with-builtin=pulse
     build-environment: *buildenv
+    override-build: |
+      ./autogen.sh
+      craftctl default
     build-packages:
       - libasound2-dev
       - libvorbis-dev
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -105,6 +105,9 @@ parts:
       - -Doptimization=3
       - -Ddebug=true
     build-environment: *buildenv
+    override-pull: |
+      craftctl default
+      patch -p1 < $CRAFT_PROJECT_DIR/patches/libffi-enable-building-on-riscv64.patch
 
   glib:
     after: [ libffi, meson-deps ]
--- a/patches/libffi-enable-building-on-riscv64.patch
+++ b/patches/libffi-enable-building-on-riscv64.patch
@@ -1,0 +1,48 @@
+From 339f603ab6b90df65de5d517ae2f692c35a9cd99 Mon Sep 17 00:00:00 2001
+From: Heinrich Schuchardt <heinrich.schuchardt@canonical.com>
+Date: Mon, 6 Feb 2023 17:28:49 +0100
+Subject: [PATCH 1/1] meson.build: enable building on riscv64
+
+The meson specific files up to now lack support for the RISC-V
+architecture.
+
+Part-of: <https://gitlab.freedesktop.org/gstreamer/meson-ports/libffi/-/merge_requests/18>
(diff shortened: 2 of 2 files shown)


snapcraft.yaml
```diff
@@ -941,6 +941,11 @@ parts:
       - --prefix=/usr
       - --with-builtin=pulse
     build-environment: *buildenv
+    override-build: |
+      set -eux
```
sergiusens: override does this already before running this, or should at least

snapcraft.yaml
```diff
@@ -941,6 +941,11 @@ parts:
...
+      cd $CRAFT_PART_BUILD
```
sergiusens: fwiw, this is the default work directory

snapcraft.yaml
```diff
@@ -941,6 +941,11 @@ parts:
...
+      ./autogen.sh --prefix=/usr
```
sergiusens: is there a difference on running this with options versus running it sans options and then running the configure script with this option?

xypron: Thanks for reviewing.

--prefix=/usr has no effect. autogen.sh only runs configure if the first argument is "c". Otherwise the argument is ignored. So I should drop it here.

snapcraft.yaml
xypron: `set -eux` is in effect without this line. I will drop this line. For consistency the same should be done for the other occurrences in a separate PR.

snapcraft.yaml
xypron: I will drop the line. For consistency the same should be done for the rest of snapcraft.yaml with a separate PR.

xypron:
@sergiusens Patch updated considering your review comments.

sergio-costas:
@xypron I dislike putting an specific commit for libffy. Either that code is integrated in a tag, or we include a patch.

xypron:
> @xypron I dislike putting an specific commit for libffy. Either that code is integrated in a tag, or we include a patch.

@sergio-costas Thanks for reviewing. Sure I can add the patch via override-pull. There are already examples in snapcraft.yaml.

sergio-costas:
@xypron Thanks!

sergio-costas:
LGTM. I'll do a test.

xypron:
Build time on a SiFive HiFive Unmatched was ca. 9 h (with `snapcraft --use-lxd`).

sergio-costas:
I'm just checking that it doesn't break in X86_64 O:-)

sergio-costas:
It seems to not break anything on x86_64, so... LGTM!

sergio-costas (approved):
LGTM

snapcraft.yaml
```diff
@@ -942,6 +945,9 @@ parts:
       - --prefix=/usr
       - --with-builtin=pulse
     build-environment: *buildenv
+    override-build: |
```
sergiusens, line 948: why is this override block now needed?

sergiusens: I did some digging, can we ensure config.guess is updated upstream?

xypron: Lennart Poettering has not released a new version of [libcanberra](http://0pointer.de/lennart/projects/libcanberra/) since 2012. This project looks orphaned.

xypron: This thread gives some background: https://discourse.gnome.org/t/is-libcanberra-deprecated/10336

kenvandine (approved):
Thanks!

PR #149, opened 2023-06-13
Some programs, like Cambalache, require Broadway

sergio-costas:
I'm creating a snap for Cambalache, but it requires Broadway https://gitlab.gnome.org/jpu/cambalache


sergio-costas:
@kenvandine built and tested with gnome-text-editor, transporter, gnome-contacts, gnome-calendar, gedit.

kenvandine (approved):
Looks good, I have no issue with enabling this

PR #151, opened 2023-06-13
Update gjs to version `1.76.2`

### LINKED ISSUE #150: Outdated gjs
The gjs version in gnome-sdk hasn't been updated for almost 5 months if not more. Which may is causing these kind of issues.
https://github.com/vixalien/muzika/issues/6

The gjs needs to be updated, to a latest one. Version `1.75.1` seems a stable one.

soumyaDghosh:
Fixes #150

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1223,9 +1223,9 @@ parts:
     build-environment: *buildenv
 
   gjs:
-    after: [ libhandy, meson-deps, gtk4 ]
+    after: [ libhandy, meson-deps ]
     source: https://gitlab.gnome.org/GNOME/gjs.git
-    source-tag: '1.76.2'
+    #source-tag: '1.76.2'
 # ext:updatesnap
 #   version-format:
 #     same-minor: true
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1225,7 +1225,7 @@ parts:
   gjs:
     after: [ libhandy, meson-deps ]
     source: https://gitlab.gnome.org/GNOME/gjs.git
-    source-tag: '1.73.2'
+    source-tag: '1.75.1'
 # ext:updatesnap
 #   version-format:
 #     same-minor: true
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1241,7 +1241,7 @@ parts:
     build-environment: *buildenv
     build-packages:
       - dbus
-      - libmozjs-91-dev
+      - libmozjs-102-dev
 
   p11-kit:
     after: [ gjs, meson-deps ]
@@ -1549,7 +1549,7 @@ parts:
       - liblzo2-dev
       - libmount-dev
       - libmount1
-      - libmozjs-91-dev
+      - libmozjs-102-dev
       - libmtdev1
       - libnettle8
       - libnghttp2-14
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1225,7 +1225,7 @@ parts:
   gjs:
     after: [ libhandy, meson-deps ]
     source: https://gitlab.gnome.org/GNOME/gjs.git
-    source-tag: '1.75.1'
+    source-tag: '1.76.1'
 # ext:updatesnap
 #   version-format:
 #     same-minor: true
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1225,7 +1225,7 @@ parts:
   gjs:
     after: [ libhandy, meson-deps ]
     source: https://gitlab.gnome.org/GNOME/gjs.git
-    source-tag: '1.76.1'
+    source-tag: '1.76.2'
 # ext:updatesnap
 #   version-format:
 #     same-minor: true
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1225,7 +1225,7 @@ parts:
   gjs:
     after: [ libhandy, meson-deps ]
     source: https://gitlab.gnome.org/GNOME/gjs.git
-    #source-tag: '1.76.2'
+    source-tag: '1.76.2'
 # ext:updatesnap
 #   version-format:
 #     same-minor: true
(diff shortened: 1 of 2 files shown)


jbicha:
1.75.1 is not a stable release (it is a pre-stable development snapshot). Maybe you're thinking of 1.76.1?

soumyaDghosh:
> 1.75.1 is not a stable release (it is a pre-stable development snapshot). Maybe you're thinking of 1.76.1?

I am trying to find that detail. Can you kindly help me? The main point of mine here is it needs to be updated.

soumyaDghosh:
https://gitlab.gnome.org/GNOME/gjs/-/commit/1abd669ae0f805a83c50738857291ae50d32f6eb

As much this suggests, should I update gjs to `1.76.1` then?

jbicha:
It might not be written down anywhere, but I know that gjs 1.75.1 is not a stable series. The part after the 1 (the 75) needs to be an even number to be considered a stable release for the gjs project. Some projects handle version numbers that way and some don't. It is confusing.

I don't know whether gjs 1.76.1 is appropriate or not, but I know that 1.75.1 is not. But gjs 1.76.1 is probably ok since we have glib 2.76.3 which was released around the same time.

sergio-costas:
At this moment we are keeping the version of GJS to 1.72.x. @kenvandine is who should approve a change like this.

sergio-costas:
Anyway, you can also compile your own GJS version and include it inside your snap.

kenvandine:
@3v1n0 suggested it's safe to update GJS, so let's give that a try.  We should go with 1.76.2 which fixed some significant memory leaks.  Can you please update that?

soumyaDghosh:
> @3v1n0 suggested it's safe to update GJS, so let's give that a try. We should go with 1.76.2 which fixed some significant memory leaks. Can you please update that?

The problem I am getting with gjs `1.76`+ is that it's showing that `have_gtk4` unknown variable. 

```
:: Configuring testNamespace.test using configuration
:: Configuring testPackage.test using configuration
:: Configuring testParamSpec.test using configuration
:: Configuring testPrint.test using configuration
:: Configuring testPromise.test using configuration
:: Configuring testRegress.test using configuration
:: Configuring testSignals.test using configuration
:: Configuring testSystem.test using configuration
:: Configuring testTweener.test using configuration
:: Configuring testCairo.test using configuration
:: Configuring testWarnLib.test using configuration
::
:: ../src/installed-tests/js/meson.build:206:3: ERROR: Unknown variable "have_gtk4".
```
`gtk4` is built before `gjs`, I guess some build-environment variables needs to be added, to add `gtk4`.

This is the snap build link

https://launchpad.net/~soumyadghosh/+snap/gnome-42-2204-sdk

sergio-costas:
Mmm... since it is for some kind of tests, maybe there is a meson option to disable building them...

soumyaDghosh:
> Mmm... since it is for some kind of tests, maybe there is a meson option to disable building them...

https://gitlab.gnome.org/GNOME/gjs/-/blob/1.76.2/installed-tests/js/meson.build?ref_type=tags#L206

If you look into this line, which is the from the `1.76.2` tag.

But, from the current master branch

https://gitlab.gnome.org/GNOME/gjs/-/blob/master/installed-tests/js/meson.build?ref_type=heads#L207

It's not there anymore.
Also `gtk_tests` are already set to `false`.

I guess we should wait for one more update? or should I build from the master branch?

soumyaDghosh:
@kenvandine @sergio-costas `gjs` from master branch with latest tag builds successfully.

snapcraft.yaml
```diff
@@ -1241,7 +1241,7 @@ parts:
     build-environment: *buildenv
     build-packages:
       - dbus
-      - libmozjs-91-dev
+      - libmozjs-102-dev
```
sergio-costas, line 1247: This package isn't available in core22 repositories.

soumyaDghosh: It's available in Jammy repos.
https://packages.ubuntu.com/jammy/libmozjs-102-dev

sergio-costas: Mmm... but I added it the snapcraft.yaml file, and snapcraft complains...

sergio-costas: Odd... now it does work.

soumyaDghosh: Should I add a patch to the `meson.build` removing the `have_gtk4` variable and update `gjs`?

kenvandine: libmozjs-102-dev is in jammy-updates and jammy-security, maybe snapcraft isn't pulling from there?

sergio-costas: I don't know what happened... I began a new build with the three changes (the GJS version and the libmozjs-102-dev), and snapcraft complained that libmozjs-102-dev didn't exist. I reverted only libmozjs, and it worked (but failed in GJS because it was the wrong libmozjs version). I changed the libmozjs package to 102 again and continued the build (this is, didn't use a `clean`) and it worked.

Now it works even after a "clean", but snapcraft was also updated in the meantime, so maybe it was a bug, or maybe it was something that I did wrong, I don't know.

sergio-costas: About the patch, maybe the best way is to check how is `master` fixing the bug instead of just removing that.

soumyaDghosh: Master removes that variable, and creates a new check which checks for it, only if `skip_gtk_tests` is set to false.

sergio-costas: Ok. Prepare the patch, then.

soumyaDghosh: Hiii, @sergio-costas @kenvandine sorry, it took me some time, but the patch for the gjs `1.76.2` is here, and seems like, it's working fine.

sergio-costas: @soumyaDghosh Thanks! I'll test it right now, but looks great.

sergio-costas: Ok, it builds perfectly, so LGTM.

sergio-costas (approved):
LGTM

PR #154, opened 2023-06-26
Add pkg-resources module

sergio-costas:
For some reason, the pkg-resources module (which is required for setuptools) wasn't being installed, so a lot of programs couldn't be snapped.

This patch fixes it.


kenvandine (approved):
Thanks!

PR #157, opened 2023-07-06
Add language-selector-common to fontconfig stage-packages

gunnarhj:
See https://github.com/snapcore/snapcraft/pull/4257#issuecomment-1622826375 for a related discussion and justification of this change.


seb128:
Seems fine to me, unless we want to be conservative on core20 and not do any of the fonts change there at this point. @kenvandine what do you think?

kenvandine (approved):
Looks good, probably worth the change.  Let's make sure we test it well.

gunnarhj:
The FF snap for jammy is [about to upgrade from core20 to core22](https://discourse.ubuntu.com/t/call-for-testing-firefox-snap-upgrading-base-from-core20-to-core22/36746). I switched to `latest/candidate/core22` on my jammy installation and noticed that the config files provided by `language-selector-common` are there, which means I can stop `~/snap/firefox/current/.config/fontconfig/fonts.conf` from loading files from the host’s `/etc/fonts/conf.d` without issues like the ones reported at https://github.com/snapcore/snapcraft/pull/4257#issuecomment-1622826375 .

So is this PR pointless now? Or is it still useful for other snaps which are still on core20?

kenvandine:
It could be useful for other snaps.

PR #159, opened 2023-08-03
Update to last versions

sergio-costas:
This patch updates to the last versions of the parts. Tested with Epiphany, drawing and shotwell.

It also fixes the python code that extracts the debug symbols: a change in a python module in pypi broke it.


kenvandine:
@sergio-costas let's split the generate_version related changes into a separate PR from the version updates.

sergio-costas:
Created a new PR with all the last updates, and separated in a different PR the tools changes.

PR #165, opened 2023-08-28
Fix strip-binaries.py

sergio-costas:
There is a problem with craft-parts, so this patch removes the dependency.


kenvandine (approved):
LGTM

PR #168, opened 2023-09-14
Update README: Describe relationship between SDK and normal snaps and point to their build farms

nteodosio:


sergio-costas (approved):
LGTM

PR #171, opened 2023-09-21
Update to last versions

sergio-costas:
It also installs all the gstream plugins

Tested with evince, drawing, gedit, zoom-client, transporter, gnome-clocks, gnome-boxes, epiphany.

Mainly it updates GLib from 2.76.4 to 2.78.0; Gtk4 from 4.10 to 4.12; json-glib from 1.6 to 1.8 (and others, updated to new revision values).


kenvandine (approved):
LGTM

PR #172, opened 2023-09-22
Add some packages to allow the use of GNOME SDK Snap as a sysroot for cross-compilation

lissyx:


kenvandine:
@sergiusens any thoughts about staging libc6 in the sdk snap to satisfy this type of use case?  We could filter it out of the content snap, so it only effects the build snap.

lissyx:
> @sergiusens any thoughts about staging libc6 in the sdk snap to satisfy this type of use case? We could filter it out of the content snap, so it only effects the build snap.

As in https://github.com/ubuntu/gnome-sdk/pull/180 ?

sergiusens:
you will need to keep this in sync with the base, I don't know if you remember that time where snapcraft crashed (or snaps that came out of it) as the base had updated glibc, but snapcraft had not, it is of low chance, but can happen if the build-snap mixes up with libraries from the base.

Make sure the gcc or python in the build snap makes use of the libc6 from this snap, perhaps consider using base bare instead.

lissyx:
> you will need to keep this in sync with the base, I don't know if you remember that time where snapcraft crashed (or snaps that came out of it) as the base had updated glibc, but snapcraft had not, it is of low chance, but can happen if the build-snap mixes up with libraries from the base.
> 
> Make sure the gcc or python in the build snap makes use of the libc6 from this snap, perhaps consider using base bare instead.

I'm sorry but I dont know if you are replying to me or thinking out loud, or what I need to do to fix this.

kenvandine:
> > you will need to keep this in sync with the base, I don't know if you remember that time where snapcraft crashed (or snaps that came out of it) as the base had updated glibc, but snapcraft had not, it is of low chance, but can happen if the build-snap mixes up with libraries from the base.
> > Make sure the gcc or python in the build snap makes use of the libc6 from this snap, perhaps consider using base bare instead.
> 
> I'm sorry but I dont know if you are replying to me or thinking out loud, or what I need to do to fix this.

It was more for me, sorry for the confusion.  We've fought with this in the past and I knew @sergiusens would remember.  I do think changing the base for gnome-42-2204-sdk to bare might be interesting if we do pull libc6 in here, perhaps something to test.  Or maybe we can start gnome-46-2404-sdk with that right from the beginning.

lissyx:
> > > you will need to keep this in sync with the base, I don't know if you remember that time where snapcraft crashed (or snaps that came out of it) as the base had updated glibc, but snapcraft had not, it is of low chance, but can happen if the build-snap mixes up with libraries from the base.
> > > Make sure the gcc or python in the build snap makes use of the libc6 from this snap, perhaps consider using base bare instead.
> > 
> > 
> > I'm sorry but I dont know if you are replying to me or thinking out loud, or what I need to do to fix this.
> 
> It was more for me, sorry for the confusion. We've fought with this in the past and I knew @sergiusens would remember. I do think changing the base for gnome-42-2204-sdk to bare might be interesting if we do pull libc6 in here, perhaps something to test. Or maybe we can start gnome-46-2404-sdk with that right from the beginning.

Being able to cross-compile non amd64 snaps would really help us on a few points (especially if it is also cross-compiled on launchpad, this would unblock some issues related to debug symbols)

lissyx:
@kenvandine @sergiusens Can we try to find a way to move this forward ?

kenvandine:
This might be a good thing to tackle when we transition to gnome-46-2404

lissyx:
So, revisiting the situation, it looks like we might anyway prefer not to use `--sysroot=` at all. I have https://github.com/canonical/firefox-snap/pull/45 and while there are currently some limitations on snapcraft, once worked around, it works pretty well.

PR #173, opened 2023-09-25
Update libadwaita to version 1.4

sergio-costas:
Tested with Drawing, gnome-calculator, gnome-text-editor and mattermost.

The patch for Yaru accent colors had to be redone because the original didn't apply anymore.


kenvandine (approved):
:+1:

PR #176, opened 2023-09-27
add libsystemd-dev to debs

jssotomdz:
Adding `libsystemd-dev` since some gnome 45 snaps are requiring it to build and this might become a build dependency for most of them as they get updated


kenvandine (approved):
Great, thanks!

PR #181, opened 2023-10-23
Added some more gstreamer related packages

soumyaDghosh:
1. Development packages for gstreamer for apps building properly
2. gstreamer-gl: for glsinkbin
3. gstreamer-pipewire: for pipewire
4. gstreamer ugly plugins for some codecs
5. gstreamer-tools: some apps use that to test the existence of gstreamer apis and plugins

One package that's necessary, but problematic, is the `gstreamer1.0-libav` package. It needs libavcodec, but fetching it will make the ffmpeg content snap and this content snap itself(probably) to clash.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1560,9 +1560,16 @@ parts:
       - libgnutls28-dev
       - libgnutls30
       - libgspell-1-dev
-      - libgstreamer-plugins-base1.0-0
-      - libgstreamer-plugins-bad1.0-0
-      - libgstreamer-plugins-good1.0-0
+      - libgstreamer-plugins-base1.0-dev
+      - libgstreamer-plugins-bad1.0-dev
+      - libgstreamer-plugins-good1.0-dev
+      - gstreamer1.0-plugins-good
+      - gstreamer1.0-plugins-bad
+      - gstreamer1.0-plugins-base
+      - gstreamer1.0-tools
+      - gstreamer1.0-plugins-ugly
+      - gstreamer1.0-pipewire
+      - gstreamer1.0-gl
       - libgudev-1.0-dev
       - libgvc6
       - libgraphviz-dev


seb128 (changes requested):
Thanks for working on that. We probably want to add gstreamer1.0-gl and -pipewire but the change as proposed increases the sdk snap from 420M to 580M. If we are adding codecs I think we need to filter the list and be more selective (and also look at the depends) and create a new content snap for the extra codecs.

Could you reduce the changes to the gl/pipewire plugins for now (and maybe the lib -> dev which seems to make sense in the sdk snap context?)

Also could you give an example of application relying on the tools? Having those in /snap/...  isn't probably to be enough by itself to make those work out of the box since the binaries aren't going to be in a path directory

soumyaDghosh:
Okay, so, as you're saying to drop the `gstreamer-gl` and `gstreamer-pipewire`, I would request you to reexamine on it. I am giving some examples on this:

1. `Identity` : This snap just needs to show image and video. To show `mp4` `mkv` and other videos `gstreamer-ugly` seems to be needed. You can try the snap. It works.
2. `Video-Trimmer` : Here comes the interesting thing with this snap. When I was successfully able to make the Identity snap work with `mp4` files, I added the same files and made the same changes with this snap. But this snap doesn't work.
3. `Decoder` : Uses gstreamer device capture for camera. But I think it needs some more changes. I was not able to make it work even with `gstreamer-pipewire`. This same config will be needed with the `Snapshot` app, the new default camera app by gnome.

I think, if these three apps works perfect, that config/`snapcraft.yaml` will be the perfect one for most of the apps.

EDIT: This forum post also helped me a lot to understand the needs.
https://forum.snapcraft.io/t/vaapi-hardware-encoding-with-gstreamer-in-snap-not-functioning-despite-successful-vainfo-output/37364

seb128:
> Okay, so, as you're saying to drop the `gstreamer-gl` and `gstreamer-pipewire`,

No, I'm suggesting to include those. And yes the situation isn't going to be perfect, some applications will require extra codecs not provided here, but increasing the size of the platform for everyone is also expensive

PR #182, opened 2023-10-25
Update to last versions

sergio-costas:
Tested with Cheese, epiphany, evince, drawing, chromium and firefox.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -169,7 +169,6 @@ parts:
       - -Dxcb=enabled
       - -Dtee=enabled
       - -Dzlib=enabled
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -169,7 +169,7 @@ parts:
       - -Dxcb=enabled
       - -Dtee=enabled
       - -Dzlib=enabled
-      - -Dgl-backend=disabled
+      #- -Dgl-backend=disabled
```
kenvandine: Please remove the line rather than commenting out

kenvandine (approved):
Thanks, looks good

PR #183, opened 2023-11-08
gtk3: gdkglcontext-wayland: Fallback to GLES 2.0 after GL failed 

kenvandine:
Apply gdkglcontext-wayland-Fallback-to-GLES-2_0-after-GL-failed.patch to fix GL and flutter apps on the Raspberry Pi 4 as well as older generation x86_64 hardware.

For reference: https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/6449


seb128 (approved):
Thanks, indeed that's needed to make flutter and GtkGL applications work on some hardware

PR #184, opened 2023-11-13
Update to last versions

sergio-costas:
Only webp-pixbuf-loader isn't updated to last version because it requires a newer version of libwebp. I'll check if it's possible to compile it, but in a different MR.

Tested with cheese, chromium, discord, drawing, epiphany, element and zoom.


kenvandine (approved):
LGTM

PR #186, opened 2023-11-23
snapcraft: Include libpipewire plus the GStreamer pipewire elements

jhenstridge:
To make use of the various pipewire-related portal APIs, we need `libpipewire`. Most will also use the `pipewiresrc` GStreamer element.

I've also explicitly staged the "base" and "good" GStreamer plugins. They were already included as as dependencies of other packages, but I think we want to be explicit about including these particular elements in the snap.

Note that to actually use libpipewire, we need to set a number of environment variables. This will require a companion PR to snapcraft-desktop-integration: https://github.com/snapcore/snapcraft-desktop-integration/pull/20

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -132,6 +132,9 @@ parts:
       - gir1.2-ggit-1.0
       - gir1.2-gucharmap-2.90
       - gir1.2-vte-2.91
+      - gstreamer1.0-plugins-base
+      - gstreamer1.0-plugins-good
+      - gstreamer1.0-pipewire
       - ibus-gtk3
       - libasound2
       - libasyncns0
@@ -177,6 +180,7 @@ parts:
       - libmtdev1
       - libogg0
       - libpathplan4
+      - libpipewire-0.3-0
       - libpng16-16
       - libpulse0
       - libpython3.10
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -132,6 +132,7 @@ parts:
       - gir1.2-ggit-1.0
       - gir1.2-gucharmap-2.90
       - gir1.2-vte-2.91
+      - gstreamer1.0-gl
       - gstreamer1.0-plugins-base
       - gstreamer1.0-plugins-good
       - gstreamer1.0-pipewire


jhenstridge:
I've also added `gstreamer1.0-gl`, since some of the GTK4 apps seem to assume it'll be there if they can create a GL context for their window.

With these changes, the platform snap is about 3 MB larger than what's published on edge. That's a bit more conservative than #181.

seb128:
Thanks, that makes sense to me. I do wonder if we should try to build a pipewire more recent than the jammy one though since the version there is quite outdated and there has been quite some work done upstream since (some might be needed for the portals to work properly?)

kenvandine:
This looks good, but I agree with @seb128, it would be a good idea to build a later version of libpipewire.  I had to build later versions format least 2 snaps recently.

jhenstridge:
The camera portal seems to function okay with the 22.04 version of pipewire. As does the screen-sharing portal, as exercised by Firefox.

One thing I have discovered is that mixing versions between libpipewire and its modules will likely lead to segfaults. This was the case when I was testing the `ashpd-demo` snap: it ships a new pipewire, but my desktop-launch changes directed it to use the modules from 22.04. Deleting pipewire from the `ashpd-demo` snap made the problem go away.

I suspect if we'd see a related crash in the Firefox snap if we were to ship new pipewire here:

* It currently bundles pipewire 0.3.48 from the Ubuntu 22.04 archive.
* It sets `PIPEWIRE_MODULE_DIR` and `SPA_PLUGIN_DIR` to point at plugins bundled in the snap via its `meta/snap.yaml`.
* It calls the gnome platform snap's `command-chain/desktop-launch` script, which can override environment set by `snap.yaml`.

So with a new pipewire in the platform snap, we'd be mixing old libpipewire with new plugins.

Unless there are known problems in the old pipewire, I'd be inclined to stick with it.

iomezk:
PipeWire 1.0.0 has been released (2023-11-26)  
https://gitlab.freedesktop.org/pipewire/pipewire/-/releases/1.0.0  
>It is API and ABI compatible with previous 0.3.x releases

seb128:
James' argument is reasonable there, should we merge the change as proposed to unblock things? We can always iterate with another change to update pipewire later if wanted

seb128 (approved):

kenvandine (approved):
LGTM

seb128:
I've merged it now so we get a build in candidate which we can try. With those changes and the environment set up from https://github.com/snapcore/snapcraft-desktop-integration/pull/20 (needs to be reviewed next) I've got ashpd-demo to work with the camera portal locally.

soumyaDghosh:
Even with this, I am unable to play videos using pipewire

```
0:00:01.398209133 223484 0x5583ff88cd20 WARN                 nvcodec gstnvrtcloader.c:116:gst_nvrtc_load_library: Could not open library libnvrtc.so, libnvrtc.so: cannot open shared object file: No such file or directory
0:00:01.446113552 223484 0x5583ff88cd20 WARN               vadisplay gstvadisplay.c:347:gst_va_display_initialize:<vadisplaydrm0> vaInitialize: unknown libva error
0:00:03.145634672 223293 0x55855d956c90 ERROR                default gstpipewirecore.c:111:make_core: (null)
0:00:03.146063917 223293 0x55855d956c90 WARN                 default gstpipewirecore.c:154:gst_pipewire_core_get: could not create core
0:00:03.146107513 223293 0x55855d956c90 WARN            pipewiresink gstpipewiresink.c:749:gst_pipewire_sink_open:<pipewiresink0> error: Failed to connect
0:00:03.146145342 223293 0x55855d956c90 ERROR                playbin gstplaybin2.c:5548:activate_group:<playbin> failed to activate sinks
0:00:03.146163616 223293 0x55855d956c90 ERROR                Clapper gstclapper.c:1169:emit_error:<clapperplayer0> Error: Failed to play (gst-clapper-error-quark, 0)
```

PR #194, opened 2024-01-30
New updates

sergio-costas:
Only webp-pixbuf-loader isn't updated to last version because it requires a newer version of libwebp. I'll check if it's possible to compile it, but in a different MR.

Tested with cheese, chromium, discord, drawing, epiphany, element and zoom.


soumyaDghosh:
@sergio-costas I have created a new PR with webp-pixbuf-loader and other updated. Here it's #198 
If it can be merged, I hope this PR can be closed.

PR #195, opened 2024-02-06
build: bump dependencies to latest upstream release

jssotomdz:
Ready to merge

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -59,7 +59,7 @@ parts:
     after: [ ninja ]
     plugin: nil
     source: https://github.com/mesonbuild/meson.git
-    source-tag: '1.2.3'
+    source-tag: '1.3.1'
     source-depth: 1
     override-build: |
       python3 -m pip install --break-system-packages .
@@ -113,7 +113,7 @@ parts:
   glib:
     after: [ libffi, meson-deps ]
     source: https://gitlab.gnome.org/GNOME/glib.git
-    source-tag: '2.78.1'
+    source-tag: '2.78.4'
 # ext:updatesnap
 #   version-format:
 #     ignore-odd-minor: true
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -898,12 +882,14 @@ parts:
       - -Dmaintainer-mode=true
       - -Dbuild-documentation=false
       - -Dbuild-demos=false
+    build-packages:
+      - graphviz
```
jssotomdz, line 886: @sergio-costas graphviz is required to build docs, and the dependency is turned on by maintainer mode (see https://gitlab.gnome.org/GNOME/gtkmm/-/blob/master/meson.build?ref_type=heads#L93) is it necessary to have maintainer mode on?

sergio-costas: Honestly, I don't know. I suspect that it is something that was already there before I arrived. @kenvandine ?

kenvandine: I don't recall, but it would be better to disable maintainer mode wherever possible

jssotomdz: by having second look at `meson.build`, maintainer mode is turned on whenever a git build is performed, which is always the case so it doesn't make a lot of sense to turn it off and hence `graphviz` is a must

kenvandine: Actually, since what we're building is really a release build from a git repo, I think it would be better to explicitly disable maintainer mode wherever possible.  What do you think @sergio-costas?

jssotomdz:
@kenvandine @sergio-costas any feedback or blockers on this PR? I'm already shifting focus towards the content snap and would like to mark this as done if that's the case

kenvandine (approved):
This looks good to me.  As I mentioned it would be nice to explicitly disable maintainer mode, since we do build from git tags and some of the parts might get that enabled.  For release builds it feels like it would be cleaner to disable that, but would like to see what @sergio-costas thinks.  That should be handled in a followup PR, if at all.

snapcraft.yaml
sergio-costas: The point is... what is "maintainer mode"? I've been checking the sources, and it seems that enabling it adds several files with C++ classes for each Gtk widget. But I presumed that they should be autogenerated from gobject introspection...

sergio-costas: Ok, according to Emanuel Bassi, the maintainer mode rebuilds the wrapper files, which I think that is a good idea...

PR #196, opened 2024-02-14
deps: add libsigc++3.0

jssotomdz:
content snap builds fine, adding `libsigc++3.0-0` since it was required in the sdk and I think it makes sense to bundle it given that the previous version is bundled

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -188,6 +188,7 @@ parts:
       - libpython3.12
       - librsvg2-2
       - libsigc++-2.0-0v5
+      - libsigc++-3.0-0
       - libsndfile1
       - libthai0
       - libtiff6


seb128 (commented):
Hey Jesus, thanks for the change. I didn't review the sdk changeset before it got merged, but one question I have ... do we need to keep libsigc++-2.0-0v5 there? It seems we are still building gtkmm3 and the requirement is coming from there? should we switch to the newer serie? 

There is an argument for keeping the gtk3 version but maybe we can also do without it and tell those who need it to rely on the core22 snap instead?

@kenvandine do you have any opinion?

jssotomdz:
@seb128 besides some gnome apps like gnome games that are still based on older versions of gnome there might be other apps requiring the library? Not sure if everything in gnome 46 will remove the dependency with libsigc++2? For some cases we bundle both old and new libraries (libwebgtk, libsoup, gtk itself) because some apps aren't ported to a newer version? Just some thoughts I had when deciding to keep the old one but I also see your point

kenvandine:
I haven't looked at it closely, but my initial thought is keeping them both would allow more apps to be rebased on core24.  But it's not unreasonable to say apps that require that lib has to stay on core22.  We will need to continue to support the gnome platform for core22, however we often see snap improvements that follow the base so those apps might miss out on those improvements.

seb128:
how much difference does it make on the snap size to not have the gtk3 bindings and corresponding libraries? if upstream follows current versions and a new base makes sense the base libraries should probably not used outdated versions either...

seb128:
Let's reject this one, there hasn't been any request for that depends to be added so we can probably do without it

PR #202, opened 2024-04-19
snap: use archive wayland, libva and intel drivers

Saviq:
These will come from the gpu-2404 provider instead (mesa-2404 by default). Any newer versions required will be handled as tracks there for hardware enablement, though they will need to remain ABI compatible with 24.04.

Ref. https://forum.snapcraft.io/t/rfc-migrating-gnome-and-kde-snapcraft-extensions-to-gpu-2404-userspace-interface/39718

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -493,22 +493,6 @@ parts:
     build-packages:
       - gtk-doc-tools
 
-  wayland:
-    after: [ librest, meson-deps ]
-    source: https://gitlab.freedesktop.org/wayland/wayland.git
-    source-tag: '1.22.0'
-    source-depth: 1
-    plugin: meson
-# ext:updatesnap
-#   version-format:
-#     no-9x-revisions: true
-    meson-parameters:
-      - --prefix=/usr
-      - -Doptimization=3
-      - -Ddebug=true
-      - -Ddocumentation=false
-    build-environment: *buildenv
-
   wayland-protocols:
     after: [ wayland, meson-deps ]
     source: https://gitlab.freedesktop.org/wayland/wayland-protocols.git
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -1605,7 +1521,9 @@ parts:
       - libudev1
       - libunity-dev
       - libuuid1
+      - libva-dev
       - libvorbisfile3
+      - libwayland-dev
```
sergio-costas, line 1493: In "Rely on gpu-2404" you remove the libraries corresponding to these packages. Is "gpu-2404" using these packages? Because if it isn't, then maybe we should rely here on the headers from gpu-2404...

Saviq: If you end up needing a newer version of libwayland than is in the gpu snap (which will, by default, be the 24.04 version), it's either the app or the GNOME snap again that will need to bring it in.

The GPU snap does not have (and shouldn't have) headers, it's not meant to be developed against, because you get the assurance that the libraries in there are 24.04 ABI-compatible. That's different from the GNOME stack because users may swap the gpu snap for another (newer version, proprietary stack etc.), so the contract has to be on the ABI level.

sergio-costas: Oh, ok... you are right.

sergio-costas (approved):
Ok, it looks fine. Since the GPU snap must remain ABI compatible, removing these and replacing them with the *-dev packages looks fine.

jssotomdz (approved):

snapcraft.yaml
soumyaDghosh: But as in the nature of development in the gnome world, they are gonna update the needed version of `wayland` and I really think it's always better to have latest version of both `wayland` and `libva`. If we eventually get to a point where we'll need to bring back the wayland, libva and other libraries, what is the use of removing those. Please kindly rethink about this.

soumyaDghosh (approved):

snapcraft.yaml
Saviq: libwayland is the less problematic one. Personally I'd keep to the archive as long as that's possible. More, I'd switch the sources for all those to the latest Ubuntu stable one so that we get our security team's advantage.

But libva we should stick to 24.04 versions. Otherwise no one can build a GPU enablement snaps that has VAAPI support that GNOME snaps would work with. The immediate example of that is Apple Silicon - for 22.04 that meant maintaining a separate track for the GNOME stack. But then there's hardware enablement, ARM stacks... we really should strive to have broad buy-in into a solution that doesn't require everyone to branch out for every hardware stack out there.

Saviq:
Anything to resolve here?

soumyaDghosh:
> Anything to resolve here?

Can you kindly shed some light on vulkan also. Will it also be provided by gpu-2404?

Saviq:
> Can you kindly shed some light on vulkan also. Will it also be provided by gpu-2404?

Yes, already is (and was for graphics-core22):

https://mir-server.io/docs/the-gpu-2404-snap-interface#heading--lists--supported-apis

Saviq:
And here, can we get this in?

seb128 (approved):
Alright, let's give that a try!

seb128:
Hum, that change is buggy, other parts are using 'after: wayland' and didn't get updated.

https://launchpadlibrarian.net/732478557/buildlog_snap_ubuntu_noble_amd64_gnome-46-2404-sdk_BUILDING.txt.gz
> A part named 'wayland' is not defined in the parts list.

Could you work on a follow-up fix?

Saviq:
Oh, sorry, coming right up.

PR #206, opened 2024-05-08
Update some components to newer series to match Noble

seb128:
I didn't test build yet but still would welcome feedback on the proposed changes. Once merged we will probably also want to do a round of tag refresh for stable serie updates

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -185,11 +185,10 @@ parts:
   gobject-introspection:
     after: [ cairo, meson-deps ]
     source: https://gitlab.gnome.org/GNOME/gobject-introspection.git
-    source-tag: '1.74.0'
+    source-tag: '1.80.0'
 # ext:updatesnap
 #   version-format:
 #     ignore-odd-minor: true
-#     lower-than: 1.76.0
     source-depth: 1
     plugin: meson
     meson-parameters:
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -269,11 +268,11 @@ parts:
   at-spi2-core:
     after: [ atk, meson-deps ]
     source: https://gitlab.gnome.org/GNOME/at-spi2-core.git
-    source-tag: 'AT_SPI2_CORE_2_44_1'
+    source-tag: '2.50.2'
```
sergio-costas: Mmm... are you sure about this?
imagen

sergio-costas: imagen

sergio-costas (changes requested):
Just one detail.

snapcraft.yaml
seb128: Thanks, I got confused by the webUI of /tags. I've changed at-spi back to the right tag format and forced pushed

sergio-costas (approved):

kenvandine (approved):

seb128:
I've pushed another iteration with extra tweaks, some notes

- gee hasn't changed much in a while and the noble version is the same so switch the stage the deb instead of building
- removing at-spi2-atk which has been merged in at-spi2-core
- atk needs libgirepository1.0-dev for Glib-2.0.gir

The parts build locally but I've been hitting 'Failed to stage: parts list the same file with different contents or permissions' errors on parts that didn't change so I think it might be a local env problem

kenvandine (approved):
This looks good to me, let's merge it and get an LP build into edge

PR #207, opened 2024-05-09
Remove the atk part it has been merged upstream in atspi

seb128:


snapcraft.yaml
```diff
@@ -222,31 +222,8 @@ parts:
...
-      PC=$CRAFT_PART_INSTALL/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/pkgconfig/atk.pc
-      sed -i 's#exec_prefix=/usr#exec_prefix=${prefix}#' $PC
-      sed -i 's#libdir=/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR#libdir=${prefix}/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR#' $PC
-      sed -i 's#includedir=/usr/include#includedir=${prefix}/include#' $PC
-    build-packages:
-      - libgirepository1.0-dev
-
   at-spi2-core:
-    after: [ atk, meson-deps ]
+    after: [ meson-deps ]
```
sergio-costas: Have you reworked the build dependencies, so now really use only what really needs, instead of building everything "linearly"? I ask this because in the old .yaml for 42, this depended on gee...

seb128: I don't understand why it was depending on gee, the Build-Depends of the deb are

```
Build-Depends: dpkg-dev (>= 1.22.5), debhelper-compat (= 13), dbus, libdbus-1-dev, dbus-broker [linux-any], libsystemd-dev [linux-any], libglib2.0-dev (>= 2.62), libxml2-dev, libx11-dev, libxtst-dev, meson (>= 0.63.0), pkgconf, gtk-doc-tools, gi-docgen, python3-sphinx, gir1.2-dbus-1.0-dev, gir1.2-glib-2.0-dev, gir1.2-gobject-2.0-dev, gobject-introspection | dh-sequence-gir, xauth <!nocheck>, xvfb <!nocheck>, at-spi2-core <!nocheck>
```

it seems to be working like that but I'm happy to change and add a depends if there is a reason for it

sergio-costas: Mmm... in that case, you should at least add glib and gobject-introspection in the "after" statement...

It was depending on gee because in the old file, each part depended on the previous one, and they were ordered in a way that ensured that, in that order, they would compile fine, although with the problem that modifying one part forced rebuilding all the later parts, no matter if they really depended on it or not.

seb128: I guess adding after glib, g-i makes sense, I did that change and force pushed

kenvandine (approved):
Ah, great fix, thanks!

PR #208, opened 2024-05-10
Enable introspection in the glib part build

seb128:
The bindings moved between those components upstream this cycle.

We are using build-packages for gobject-introspection instead of using the local part to avoid a dependency loop, we could also build glib once without introspection/don't stage that one but it seems that if the distro g-i version is recent enough it's simpler to just use it

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -122,6 +122,7 @@ parts:
       - --prefix=/usr
       - -Doptimization=3
       - -Ddebug=true
+      - -Dintrospection=enabled
     build-environment: *buildenv
     override-build: |
       set -eux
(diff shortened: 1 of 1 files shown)


seb128:
I've not build tested it yet but I'm doing that now

For extra context
https://gitlab.gnome.org/GNOME/gobject-introspection/-/commit/afbe7697e

```
build: Do not install generated GLib GIR files

GLib 2.79 ships its own introspection data.

We need to keep generating the introspection data for GIRepository-2.0,
but we don't need to install its dependencies.
```

kenvandine (approved):
Thanks, this should be safe as we ensure what's staged comes first in XDG_DATA_DIRS so following parts should use the build Gir where possible

PR #211, opened 2024-05-10
Don't use virtual packages for the build

seb128:
Resolve those warnings from the current build

```
Installing build-packages
libfreetype6-dev is a virtual package, use non-virtual packages for deterministic results.
libidn2-0-dev is a virtual package, use non-virtual packages for deterministic results.
```


kenvandine (approved):
Nice catch, thanks

PR #217, opened 2024-05-29
ci: add build test

jssotomdz:
I realized build-test isn't part of the ci for the sdk. Is there a reason for that? This is useful to detect regressions.
Adding it to the content snap is a challenge because it depends on the sdk

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,19 @@
+name: build-test
+on:
+  pull_request:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


seb128 (approved):
It makes sense but let's wait for Ken input before merging in case there was a reason it was not enabled (could be CI delay or quota problem for example). In practice it would have caught some of the recent build regressions we had when landing changes

kenvandine (approved):
Excellent, thanks!

PR #218, opened 2024-05-29
ci: add build test

jssotomdz:
I realized build-test isn't part of the ci for the sdk. Is there a reason for that? This is useful to detect regressions.
Adding it to the content snap is a challenge because it depends on the sdk


kenvandine (approved):
Excellent, thanks!

PR #220, opened 2024-05-29
clutter: depend on json-glib

Saviq:


sergio-costas:
Mmm... but that is already in upstream...
imagen

Saviq:
It is since https://github.com/ubuntu/gnome-sdk/commit/2c8ffadb2fce9c6802924a31df8b33094d021c4c :)

PR #221, opened 2024-06-06
deps: fix libsoup3 dependency issue and bump dependencies

jssotomdz:


kenvandine (commented):
Is the jump from libsoup3 3.4 to 3.5 anything to be concerned about?

jssotomdz:
> Is the jump from libsoup3 3.4 to 3.5 anything to be concerned about?

[Changelog](https://gitlab.gnome.org/GNOME/libsoup/-/blob/master/NEWS) doesn't list anything potentially disruptive afaik

seb128 (changes requested):
Thanks, a few notes

- following Ken's question, according to https://gitlab.gnome.org/GNOME/libsoup/-/tags 3.5.1 is labelled as '(dev)' so I think libsoup probably follows the even=stable, odd=unstable versioning scheme, in which case I think we should stick to 3.4.x since we want a stable sdk be on stable series right?

- could you split the logical changes in different commits? one to fix the libsoup depends and one to bump dependency

- one thing I was wondering before and which I'm going to take the opportunity to ask (could be rather one for @kenvandine though) ... should we update poppler versions in the sdk? Upstream is bumping the soname at every release even if there is no ABI change, but would that break other snaps which are linked to the old soname?

- could you document the reason libgraphene-1.0-dev was added as a build-depends of gtk

- could you give some context on staging libigdgmm-dev?

jssotomdz:
> * could you document the reason libgraphene-1.0-dev was added as a build-depends of gtk

gtk4 was failing to build in its newest checkout tag due to `Couldn't find include 'Graphene-1.0.gir'`. Adding that as a build-depends solved the issue.

> * could you give some context on staging libigdgmm-dev?

Similar case here but a bit surprising. Line 1571 has a `sed -i 's#includedir=${prefix}/snap/gnome-46-2404-sdk/current#includedir=${prefix}#' $CRAFT_PRIME/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/pkgconfig/igdgmm.pc` which failed stating that `igdgmm.pc` doesn't exist, and it's provided by `libigdgmm-dev`.

I'll address the other comments and wait for input on the poppler question

kenvandine:
> Thanks, a few notes
> 
>     * following Ken's question, according to https://gitlab.gnome.org/GNOME/libsoup/-/tags 3.5.1 is labelled as '(dev)' so I think libsoup probably follows the even=stable, odd=unstable versioning scheme, in which case I think we should stick to 3.4.x since we want a stable sdk be on stable series right?
> 
>     * could you split the logical changes in different commits? one to fix the libsoup depends and one to bump dependency
> 
>     * one thing I was wondering before and which I'm going to take the opportunity to ask (could be rather one for @kenvandine though) ... should we update poppler versions in the sdk? Upstream is bumping the soname at every release even if there is no ABI change, but would that break other snaps which are linked to the old soname?

In practice we haven't seen an issue with that in the past, but that doesn't mean it's not an issue.  Actually maybe this is a good opportunity to drop this source part and stage the package?  It is something I would feel better about relying on the distro for from a security perspective.  IIRC we moved to building it from source because evince needed it.
> 
>     * could you document the reason libgraphene-1.0-dev was added as a build-depends of gtk
> 
>     * could you give some context on staging libigdgmm-dev?

seb128:
@jssotomdz could you change poppler to use a stage package instead of building it from source?

jssotomdz:
> @jssotomdz could you change poppler to use a stage package instead of building it from source?

done! added glib and gir packages as well since the compilation flags were turned on in the part

seb128:
@jssotomdz thanks, it's looking good, the only remaining comment was the libsoup version one, could we roll back to the stable serie (3.4)?

jssotomdz:
> @jssotomdz thanks, it's looking good, the only remaining comment was the libsoup version one, could we roll back to the stable serie (3.4)?

done! thought I had already handled it, I think I forgot to push

seb128 (approved):
Thanks!

PR #222, opened 2024-06-19
Don't stage libigdgmm-dev, it's not required and doesn't exist on armhf

seb128:
It was added as part of !221 and it breaking arm builds since the package is intel specific. Afaik the sdk was building before so it shouldn't be really required?

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1481,11 +1481,12 @@ parts:
       - zlib1g-dev
       - zlib1g
       - libwebp-dev
-      - libigdgmm-dev
       - libpoppler-dev
       - libpoppler-glib-dev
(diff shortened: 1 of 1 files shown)


seb128:
As the CI is showing that needs to be fixed, I'm going to rework it

jssotomdz:
Just a thought, I added the ci for amd64 and now an issue appears on a different architecture. Anything we can do ci-wise to test other archs?

jssotomdz (approved):

seb128:
> Just a thought, I added the ci for amd64 and now an issue appears on a different architecture. Anything we can do ci-wise to test other archs?

I'm unsure, I think github CI doesn't provide other architectures...

PR #226, opened 2024-07-04
deps: bump dependencies

jssotomdz:
special attention on gtkmm which was limited to v4 but now some apps (system monitor for now) require the new version

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -112,7 +112,7 @@ parts:
   glib:
     after: [ libffi, meson-deps ]
     source: https://gitlab.gnome.org/GNOME/glib.git
-    source-tag: '2.80.2'
+    source-tag: '2.80.3'
 # ext:updatesnap
 #   version-format:
 #     ignore-odd-minor: true
@@ -250,7 +250,7 @@ parts:
   fribidi:
     after: [ meson-deps ]
     source: https://github.com/fribidi/fribidi.git
-    source-tag: 'v1.0.14'
+    source-tag: 'v1.0.15'
     source-depth: 1
     plugin: meson
     meson-parameters:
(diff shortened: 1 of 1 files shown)


sergio-costas (approved):
Looks good, although I would also check if atkmm can be updated too, because the limit version there was due to gtkmm.

jssotomdz:
> I would also check if atkmm can be updated too

[looks](https://gitlab.gnome.org/GNOME/atkmm/-/tags) up to date

sergio-costas:
Mmm... odd... then, why is there that "2.53.3" version limit...?

seb128:
> Mmm... odd... then, why is there that "2.53.3" version limit...?

Seems like to workaround a bug in the upstream repository? Current version is 2.36 but

[c6ecc83d](https://gitlab.gnome.org/GNOME/atkmm/-/commit/c6ecc83d3b3b3087df465813064c06ecf3d0b92e) · [2.53.3](https://gitlab.gnome.org/GNOME/atkmm/-/commit/c6ecc83d3b3b3087df465813064c06ecf3d0b92e) · 7 years ago

snapcraft.yaml
```diff
@@ -294,14 +294,14 @@ parts:
...
-    source-tag: '1.52.2'
+    source-tag: '1.54.0'
     source-depth: 1
     plugin: meson
     meson-parameters:
       - --prefix=/usr
       - -Doptimization=3
       - -Ddebug=true
-      - -Dinstall-tests=false
+      - -Dbuild-testsuite=false
```
seb128, line 304: The change was https://gitlab.gnome.org/GNOME/pango/-/commit/927abcc which also added -Dbuild-examples, could you set that one to false?

seb128 (changes requested):
Thanks. I've one small tweak ask but it's looking good otherwise

snapcraft.yaml
jssotomdz: addressed!

seb128 (approved):
Thanks!

PR #227, opened 2024-07-04
deps: stage libpolkit-gobject

jssotomdz:


jssotomdz:
Wondering if I should open this same pull request against `gnome-46-2404-sdk`

seb128 (commented):
Thanks. Should we also stage the library or only the dev? (the current list seems a bit inconsistent but the dev should depends on the library)

For 46 the bug doesn't exist

```
$ find /snap/gnome-42-2204-sdk/258/ -name libpolkit*
/snap/gnome-42-2204-sdk/current/usr/lib/x86_64-linux-gnu/libpolkit-gobject-1.a
/snap/gnome-42-2204-sdk/current/usr/lib/x86_64-linux-gnu/libpolkit-gobject-1.so
```

vs

```
$ find /snap/gnome-46-2404-sdk/30/ -name libpolkit*
$
```

But it might still make sense to provide that library

jssotomdz:
> the dev should depends on the library

[Yes](https://packages.ubuntu.com/jammy/libpolkit-gobject-1-dev), you're right, depends on the library. Removing it

seb128 (approved):
Thanks

PR #229, opened 2024-07-09
Clear execstack, necessary for armhf builds, triggers review warning

kenvandine:


jhenstridge (approved):
Context from discussions elsewhere: the libde265 library is marked as having an executable stack on armhf. This is probably due to the ARM assembly code in the library. There seems to have been an attempt to mark the code safe here, but for whatever reason it hasn't taken:

https://github.com/strukturag/libde265/blob/master/libde265/arm/asm.S#L62

On the x86 side is using SSE intrinsics in C code rather than assembly, so that doesn't really offer any clues.

Manually marking the library safe seems like a reasonable way of dealing with the automated review failure. Worst case is it causes apps using the library to segfault on armhf if it indeed tries to execute code off the stack.

PR #230, opened 2024-07-09
Updated some package names that changed between jammy and noble

kenvandine:
Updated some package names that changed between jammy and noble, breaking armhf builds

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -173,11 +173,11 @@ parts:
       - libjbig0
       - libjpeg-turbo8
       - liblcms2-2
-      - libllvm17
+      - libllvm17t64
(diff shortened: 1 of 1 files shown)


seb128 (approved):
Thanks

seb128 (changes requested):
In fact those also need to be updated

libasound2 -> libasound2t64
libcups2 -> libcups2t64
libdb5.3 -> libdb5.3t64
libflac12 -> libflac12t64
libmozjs-115-0 -> libmozjs-115-0t64
libpipewire-0.3-0 -> libpipewire-0.3-0t64
libpython3.12 -> libpython3.12t64

seb128 (approved):
Thanks!

PR #237, opened 2024-08-27
Include more mesa and libdrm files

tobhe:
I've run into issues with the latest mesa update. In this case it was missing libdrm files but I found more that probably should be included too.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -45,6 +45,7 @@ parts:
     stage:
       # libegl-mesa0
       - -usr/lib/$CRAFT_ARCH_TRIPLET/libEGL_mesa.so.0.0.0
+      - -usr/share/glvnd/egl_vendor.d
       # libgbm1
       - -usr/lib/$CRAFT_ARCH_TRIPLET/libgbm.so.1.0.0
       # libglapi-mesa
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -56,21 +57,51 @@ parts:
...
       - libglapi-mesa
       - libglu1-mesa
       - libglx-mesa0
-      - libgl1-mesa-dri
       - mesa-va-drivers
+      - mesa-vulkan-drivers
+      # libdrm
+      - libdrm2
+      - libdrm-common
+      - libdrm-amdgpu1
```
kenvandine, line 96: Is amd and nouveau really relevant for the asahi track? Or is this just for consistency?

kenvandine (commented):
Is this mostly just replacing more files staged from gnome-42-2204 with those from the debs?

snapcraft.yaml
tobhe: That's the fun part: it is. the dri drivers are all a single binary built against libdrm from the ppa. If we don't also include amdgpu we get:
```
libEGL warning: MESA-LOADER: failed to open asahi: /snap/supertuxkart/677/gnome-platform/usr/lib/aarch64-linux-gnu/dri/asahi_[dri.so](http://dri.so/): undefined symbol: amdgpu_va_get_start_addr (search paths /snap/supertuxkart/677/gnome-platform/usr/lib/aarch64-linux-gnu/dri, suffix _dri)
```

tobhe:
> Is this mostly just replacing more files staged from gnome-42-2204 with those from the debs?

Right. In fact that is all the asahi snap does, but this adds a few more files from libdrm and mesa.

kenvandine (approved):
Thanks, this looks good to me

Issue #225, opened 2024-07-03
SDK contains broken link to libpolkit-gobject-1.so.0.0.0


robert-ancell:
The SDK contains the file `/snap/gnome-42-2204-sdk/current/usr/lib/x86_64-linux-gnu/libpolkit-gobject-1.so` which points to `libpolkit-gobject-1.so.0.0.0` which is not in this snap. This file is in `/snap/core22/current/usr/lib/x86_64-linux-gnu/libpolkit-gobject-1.so.0.0.0`. Since the file is missing, a program (e.g. gnome-software) will instead try to link against the static library.

I suspect the symlink needs to point to the version in the core22 snap or be removed.

nteodosio:
I try a build of the gnome-software snap with `snapcraft --verbosity=trace --debug --use-lxd`, it tries to link statically and fails.

## Remove the .so link

In the shell I remount the Gnome SDK rw and delete the symlink:

```
snapcraft-snap-store-on-amd64-for-amd64-11976706 ../project# umount /snap/gnome-42-2204-sdk/current
snapcraft-snap-store-on-amd64-for-amd64-11976706 ../project# unsquashfs -d /tmp/gnome /var/lib/snapd/snaps/gnome-42-2204-sdk_258.snap
Parallel unsquashfs: Using 8 processors
34743 inodes (45241 blocks) to write

[===========================================================================================================================================================================-] 45241/45241 100%

created 27252 files
created 2509 directories
created 7433 symlinks
created 0 devices
created 0 fifos
created 0 sockets
snapcraft-snap-store-on-amd64-for-amd64-11976706 ../project# mount --bind /tmp/gnome/ /snap//gnome-42-2204-sdk/current
snapcraft-snap-store-on-amd64-for-amd64-11976706 ../project# rm /snap/gnome-42-2204-sdk/current/usr/lib/x86_64-linux-gnu/libpolkit-gobject-1.so
```

and resume the build:

```
snapcraft-snap-store-on-amd64-for-amd64-11976706 ../project# snapcraft --verbosity=trace
```

which fails again on same error:

```
2024-09-11 11:10:43.947 :: /snap/gnome-42-2204-sdk/current/usr/bin/ld: /snap/gnome-42-2204-sdk/current/usr/lib/x86_64-linux-gnu/libpolkit-gobject-1.a(libpolkit_gobject_1_la-polkitauthority.o): warning: relocation against `polkit_authority_enumerate_actions' in read-only section `.text'
[...]

PR #238, opened 2024-09-10
Add libpolkit-gobject-1-0

nteodosio:
Otherwise libpolkit-gobject-1.so is a dangling link to libpolkit-gobject-1.so.0.0.0.

Expands #227.

Rationale
---------

If the package (libpolkit-gobject-1-0) is not present in the stage-packages it will not be included in the snap even if it is a dependency of a stage-package (libpolkit-gobject-1-dev). Also the [build log of the gnome-sdk](https://launchpadlibrarian.net/738931725/buildlog_snap_ubuntu_jammy_amd64_gnome-42-2204-sdk_BUILDING.txt.gz) makes me wonder if it is being pulled at all, as I see no reference to the *-0, only to the *-dev in it.

Bug
---

Gnome software snap [attempts to link this library statically instead of dynamically](https://code.launchpad.net/~desktop-snappers/+snap/snap-store-stable/+build/2583953/+files/buildlog_snap_ubuntu_jammy_amd64_snap-store-stable_BUILDING.txt.gz) because the .so is a dangling link while the .a is the real thing.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1653,6 +1653,7 @@ parts:
       - shared-mime-info
       - zlib1g-dev
       - libwebp-dev
+      - libpolkit-gobject-1-0
       - libpolkit-gobject-1-dev
     override-build: |
       set -eux


nteodosio:
> Gnome software snap [attempts to link this library statically instead of dynamically](https://code.launchpad.net/~desktop-snappers/+snap/snap-store-stable/+build/2583953/+files/buildlog_snap_ubuntu_jammy_amd64_snap-store-stable_BUILDING.txt.gz) because the .so is a dangling link while the .a is the real thing.

Nah, on second thought that is balderdash. It does leave me wondering why the -dev packages include a dangling link though.

This is just the SDK right, for the runtime it will be in the non-SDK snap.

nteodosio:
In my previous comment I had the erroneous assumption that at link time the library itself would not need to be there for a dynamic link, but I was wrong per https://stackoverflow.com/questions/9688200/difference-between-shared-objects-so-static-libraries-a-and-dlls-so:

> at link time the linker looks inside the library itself to make sure the functions it needs are actually there

I did not catch that at https://github.com/ubuntu/gnome-sdk/issues/225#issuecomment-2343121929 likely because at the time I provide the actual shared object meson has already configured the build.

So I re-run that test case but putting the libpolkit-gobject shared object before starting the build, rather at the end of the pull phase:

1. snapcraft pull --verbosity=trace --shell-after --use-lxd
2. Substitute libpolkit-gobject*.so* in the /snap/gnome-42-2204-sdk
3. Continue the build with snapcraft --verbosity-trace

And it built fine.

So I'm reopening this merge request.

seb128 (approved):
Thanks, the fix makes sense to me. I'm tagging Ken and Jesus also though because I lack the context of why the -dev was added and if there was any reason at the time for letting the library out (it's also unclear why to me why it's not pulled in as a depends of the dev binary)

kenvandine (approved):
This looks good to me. It wasn't intentionally left out, I would have expected it to be staged as a dependency of the dev package.

jssotomdz:
As Ken pointed out, the library is supposed to pull the library as a dependency. For some reason I still don't understand, the .so is deleted at some point. While it worked for me locally (I saw the link working) when I sent it to snapcraft the library doesn't exist (broken link). Snapcraft team told me it was because the core22 base snap is cached in my system. But besides that I don't have any further insights.

nteodosio:
Is there any way to confirming the fix apart from releasing to stable and rebuilding affected software? Build-snaps have the snap/channel format but [extensions](https://snapcraft.io/docs/snapcraft-extensions) do not.

PR #239, opened 2024-09-12
allow ci reruns and update upload artifact action

jssotomdz:


jssotomdz:
@sergio-costas @seb128 should I commit the v4 change directly into the rest of the repos? to avoid spamming and creating >20 pull requests

sergio-costas (approved):
I don't know about the "workflow_dispatch:" entry, but the version change is exactly what it has to be done to fix it.

sergio-costas:
@jssotomdz In my opinion, yes, you should do that. It's a sensible change.

seb128 (approved):
> I don't know about the "workflow_dispatch:" entry, but the version change is exactly what it has to be done to fix it.

The  workflow_dispatch gives you the ability to trigger the job manually

The changes makes sense to me, thanks!

sergio-costas:
@seb128 Oh, good to know. Thanks!

PR #241, opened 2024-09-19
Add missing dependency and a test for dependencies

sergio-costas:
The conditioning stage wasn't dependent on librest, so it could be done before, thus not "fixing" the .pc files of librest. Also, it lacked dependency on at-spi2-core and webp-pixbuf-loader.

This PR fixes this and also adds a github test to ensure that any new PR with new parts guarantee that DEBS depend on it.


jssotomdz (approved):

seb128 (approved):
Thanks, that seems fine to me

PR #242, opened 2024-09-19
Add missing dependency and tool for checking in Gnome-42

sergio-costas:
There is a missing dependency in DEBS part: webp-pixbuf-loader. This patch adds it and also adds a check to ensure that any push fulfills the dependencies, ensuring that DEBS is ran after all the previous parts.


seb128 (approved):
Thanks

PR #244, opened 2024-10-03
[gnome-42-2204] Update libadwaita and add new yaru accent colors

3v1n0:
See commits for details.

UDENG-4597


seb128 (commented):
The patches seems reasonable. Do we need the libadwaita version update? Reading https://gitlab.gnome.org/GNOME/libadwaita/-/blob/libadwaita-1-4/NEWS there are quite some changes, we will need extra checking of existing snaps if we take the new version, a more selected fix might be a better option if we want to land that fix in Oracular?

3v1n0:
It's a stable update and we've been testing those fixes for long time in noble already, so also checking the code I'm quite confident that they don't break anything, but actually improve various things (including various crash or introspection fixes).

So I might have cherry-picked them, but I believe it's just safer to trust on upstream testing that has been done for some long time now.

sergio-costas (approved):
Looks sensible.

seb128:
Alright, let's give it a try

PR #247, opened 2024-10-12
Add gcin-gtk3-immodule

ifurther:
Add gcin-gtk3-immodule for Chinese input

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -59,6 +59,7 @@ parts:
       - gir1.2-gucharmap-2.90
       - gir1.2-vte-2.91
       - ibus-gtk3
+      - gcin-gtk3-immodule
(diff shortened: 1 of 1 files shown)


sergio-costas (approved):
Looks sensible.

ifurther:
@sergio-costas Will the  pr merge or need another reviewer ?

sergio-costas:
Sorry for the delay. We found that this package isn't neither in gnome-42-2204 nor in gnome-46-2404. Can you give us an use case and describe us what doesn't work without this, that would work after installing this?

ifurther:
The input method in gtk3 need the immodule, such like the ibus and fcitx. In Firefox with snap will not work, and Firefox with deb version will work.

PR #248, opened 2024-10-14
Use .deb libtool instead of savannah version

sergio-costas:
This is to avoid the downloading problems we are having from the Savannah git repo from GNU.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -72,17 +72,13 @@ parts:
       - python3-pip
 
   libtool:
-    source: https://git.savannah.gnu.org/git/libtool.git
-    source-tag: 'v2.4.7'
-    plugin: autotools
+    plugin: nil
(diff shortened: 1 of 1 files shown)


3v1n0:
I feel we can just remove the commented code, I feel this is one of those components that taking from the distro is just better than compiling it again in the sdk, being it quite core anyways

seb128 (changes requested):
I agree with Marco there, let's just use the deb going forward

sergio-costas:
@seb128 @3v1n0 Commented code removed. Also found that it is using an old .deb for mozjs... preparing a new patch.

3v1n0:
Thanks, squash the two commits please though.

sergio-costas:
Squashed

seb128 (approved):
Thanks

PR #249, opened 2024-10-14
Update libmozjs to 115

sergio-costas:
GJS is built with libmozjs-115-dev, but in the dependencies for the SDK itself the old limbozjs-102-dev is used. This PR fixes this.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1404,7 +1404,7 @@ parts:
       - liblzma5
       - libmount-dev
       - libmount1
-      - libmozjs-102-dev
+      - libmozjs-115-dev
       - libmtdev1t64
       - libnettle8t64
       - libnghttp2-14


sergio-costas:
Should we keep both versions? (102 and 115) Or it should be enough with the latest one?

3v1n0:
Si el target es 2404 creo que sea mejor dejar solo la 115. 102 era solo para 22.04.

3v1n0:
Ops, sorry... I saw Sergio and I went Spanish :joy:.

Anyways, I think 115 is well enough for apps targetting 24.04. :)

sergio-costas:
I'm fully de acuerdo :-D

seb128 (approved):
Thanks!

3v1n0:
Oh, the only thing... On the non-sdk snap we should likely keep the old version, as the apps compiled with the old version of SDK won't work otherwise.

sergio-costas:
Mmm... but the non-sdk snap is built from the sdk one, so then we must include it here...

sergio-costas:
I'll prepare a new PR.

seb128:
I think we should remove 102, we didn't really migrate anything to gnome46 yet since we hit snapd issues trying to do

PR #250, opened 2024-10-14
Updated all parts to the last stable version

sergio-costas:


sergio-costas:
Should gnome-desktop be updated to 46.X version?

jssotomdz:
> Should gnome-desktop be updated to 46.X version?

I'd say yes, is it not being picked up by the automation because of the same-major parameter?

soumyaDghosh:
@sergio-costas @jssotomdz Can't libadwaita be updated to `1.6`? It's in stable now. Also the patches for libadwaita needs to be updated.

sergio-costas:
@soumyaDghosh The problem with libadwaita is that it requires some patches, and for version 1.6 they require some extra work. For 1.5.5, instead, just with some little adjustments they work.  So, for the moment, I will just update to 1.5.5.

soumyaDghosh:
> @soumyaDghosh The problem with libadwaita is that it requires some patches, and for version 1.6 they require some extra work. For 1.5.5, instead, just with some little adjustments they work. So, for the moment, I will just update to 1.5.5.

I believe the patches can be taken from here

https://salsa.debian.org/gnome-team/libadwaita/-/tree/ubuntu/1.6.0-1ubuntu2/debian/patches/ubuntu?ref_type=tags

These were patched by @3v1n0

seb128:
The commits are a bit confusing, could you rebase/squash as appropriate?

sergio-costas:
@seb128 Done.

But it still has to be decided whether gnome-desktop should be updated to version 46, or kept in version 44. Also, maybe you prefer to not update librsvg, since it requires a rust version bigger than the currently available as .debs. Finally, what should we do with libadwaita? Update it to 1.6.x, or keep it in 1.5.x?

seb128:
I would update gnome-desktop and libadwaita unless there is a reason to not do so?

sergio-costas:
Funny... there's no gnome-desktop 46, it's still at 44.1. So we can forget that.

sergio-costas:
Updated everything, and tested the generated "gnome-46-2404" with evince and epiphany (in the later, went to Youtube and played a video). Everything seems to work fine.

kenvandine (approved):
This looks good to me. Unrelated to your change I noticed the gtk4 part does have -Dbuild-tests=false twice, we should clean that up but not a blocker for merging this.

PR #251, opened 2024-10-16
Delete broken links

sergio-costas:
After the "mesa cleanup", libraries are removed but not some soft links to them, so they remain, but are broken.

This patch removes all the broken links.


seb128 (approved):
Thanks!

PR #252, opened 2024-10-25
fix: libsrvg needs curl to fetch rust

kenvandine:
Added curl to build-packages for librsvg as it needs to use curl to setup rust


sergio-costas (approved):
LGTM.

seb128 (approved):
Thanks

PR #253, opened 2024-11-07
Remove libsoup2 support

sergio-costas:
Three libraries (librest, libgeocode and libgweather) still used libsoup-2, although they already support being compiled with libsoup-3. Since mixing both versions when linking binaries is not possible (and, in fact, an specific check is done at runtime to abort the execution if that case is detected), I think that is time to remove the old one.

This PR entirely removes libsoup-2 library and recompiles those three libraries using libsoup-3.

I will mark it as draft to ensure that a discussion is done before merging it.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -416,29 +416,8 @@ parts:
       craftctl default
       sed -i 's#^\#!/usr/bin/env python$#\#!/usr/bin/env python3#g' src/psl-make-dafsa
 
-  libsoup2:
-    after: [ libpsl, meson-deps ]
-    source: https://gitlab.gnome.org/GNOME/libsoup.git
-    source-tag: '2.74.3'
-# ext:updatesnap
-#   version-format:
-#     lower-than: 3
-    source-depth: 1
-    plugin: meson
-    meson-parameters:
-      - --prefix=/usr
-      - -Doptimization=3
-      - -Ddebug=true
-      - -Dvapi=enabled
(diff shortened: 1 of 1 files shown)


seb128:
I think it makes sense to go forward with the new version in 46.

@jbicha you followed the upstream work and handled most of the Debian/Ubuntu archive changes, does that seem reasonable to you that we default to libsoup3 now for the new sdk?

jbicha:
Basically the only thing in the GNOME world that still uses libsoup2 is libgdata which is used by gvfs. If libgdata isn't used in this snap, then it's a good thing to stop using libsoup2. Even if libgdata is used, libsoup2 should not be used anywhere else in this snap.

sergio-costas:
It's even worse: it seems that libgdata is used only in gvfsd-google, nothing else...

sergio-costas:
There is a draft patch to port it to libsoup3, in fact, but has been frozen for a long time https://gitlab.gnome.org/GNOME/gvfs/-/merge_requests/147#note_1519119

jbicha:
Is gvfsd-google part of this snap? If it isn't, I don't think there any blockers to landing this change.

sergio-costas:
No, it isn't.

jbicha (approved):

sergio-costas:
Merging then.

seb128 (commented):
+1 from me for the change

sergio-costas:
Ops... sorry, I presumed that your vote was already positive.

PR #254, opened 2024-11-07
Update the gdk-pixbuf loaders

sergio-costas:
Currently, the `loaders.cache` file in the `gdk-pixbuf` folder contains wrong paths, because it is generated during the `sdk` build phase. This is not a problem for snaps making use of gnome-XX-YYYY because the extensions script defines two environment variables and regenerates a new `loaders.cache`file. But in the case of core-desktop this approach seems to not work, so this patch ensures that the default `loaders.cache` file is correct when `layout` is used to map the gdk-pixbuf-2.0/2.10.0 folder in the root filesystem.


jssotomdz (approved):

kenvandine (approved):
Good catch, this looks good to me

seb128 (approved):
Thanks

PR #255, opened 2024-11-14
Don't delete index.theme files in icons folder

sergio-costas:
The index.theme file is required to regenerate the icon cache, so it must be preserved when checking for duplicated files.

This patch does this, thus allowing to regenerate the icons cache after duplicates have been removed, thus ensuring that only really-available icons will be added to the cache.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -260,11 +260,7 @@ parts:
       craftctl default
       $CRAFT_STAGE/usr/bin/glib-compile-schemas $CRAFT_STAGE/usr/share/glib-2.0/schemas
       $CRAFT_STAGE/usr/bin/update-mime-database $CRAFT_STAGE/usr/share/mime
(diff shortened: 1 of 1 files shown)


seb128:
Could you give a bit more details in the impact? Is there any user visible issue in one of our snaps created by that code today?

sergio-costas:
Not in any current, but I'm working on using gnome-46 as a base snap for ubuntu-desktop-session.snap, and I'm worried that the lack of it can have impact. It is true that the index-theme.cache files are still there, but... well... for completeness.

sergio-costas:
Oh... also, I'm recreating the cache AFTER removing the duplicated icons. This is how it should be done. And that's why I must preserve the index.theme file. Without it, it's not possible to recreate it.

sergio-costas:
Maybe I should modify the title...

seb128 (approved):
Ok, that seems to make sense indeed!

PR #259, opened 2024-11-20
Restore libsoup-2.4

sergio-costas:
This PR restores libsoup-2.4 and includes two versions of libgeoclue: one compiled against libsoup-2.4 and another one against libsoup-3.0. This should mitigate the migration.


sergio-costas:
Closing because libsoup-2.4 must be added inside any snap that still requires it.

Issue #260, opened 2024-11-20
glib-mkenums fails


sergio-costas:
glib-mkenums fails to run because the shebang is `#!/usr/bin/python33`. The reason seems to be this line in `snapcraft.yaml`:

    sed -i 's#/usr/bin/python#python3#' usr/bin/glib-mkenums

I'm preparing a patch for it. After I test it, I'll upload it.

PR #261, opened 2024-11-20
Fix glib-mkenums

sergio-costas:
The 'conditioning' part is replacing '/usr/bin/python' in glib- mkenums file with '/usr/bin/python3', to ensure that it works with the current builds. Unfortunately, this only works if the original shebang was referring to python2. The current version of glib-mkenums, instead, has a shebang that already points to python3, so the result of the sed command is a shebang that points to python33, resulting in an execution error.

This patch fixes it by ensuring that the replacement is done only if the shebang points to "python", but not if it points to "python3".

Fix https://github.com/ubuntu/gnome-sdk/issues/260

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1547,7 +1547,7 @@ parts:
         sed -i 's#includedir=${prefix}/snap/gnome-46-2404-sdk/current#includedir=${prefix}#' $CRAFT_PRIME/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/pkgconfig/igdgmm.pc
       fi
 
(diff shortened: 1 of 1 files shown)


seb128 (changes requested):
Thanks. The existing code seems weird, basically it is making the interpreter be "#!python3" which isn't valid right?

Could we simply remove that sed call instead?

We know that glib is using python3 in noble and it will not go back, is there really a case where it would ever do something?

sergio-costas:
Yes, you are right. Also, I missed the fact that it removed the `/usr/bin/` part, which is completely wrong.

Done the removal. But, just in case, wait until I test it.

sergio-costas:
Confirmed: it works.

seb128 (approved):
Great, thanks!

PR #262, opened 2024-11-21
Disable all tests and documentation

sergio-costas:
This PR disables, during the building of parts, all the tests and documentation build, and also enables building gobject introspection data and VAPI files, to ensure that all libraries are available in any language.

It also removes the usr/share/gtk-doc folder, since it's not needed.

All this saves a little bit of space and slightly reduces the build time.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -149,6 +149,8 @@ parts:
       - -Doptimization=3
       - -Ddebug=true
       - -Dgtk=disabled
+      - -Dtests=disabled
+      - -Ddemos=disabled
     build-environment: *buildenv
 
   cairo:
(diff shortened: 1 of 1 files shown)


kenvandine:
This disables building and running the tests right? It's not about installing the tests? I think I'd prefer running the tests anywhere we can just so we can get build failures when an expected passing tests fails.

sergio-costas:
But, AFAIK, those tests are built but not run...

seb128 (approved):
I agree with Ken that running the tests would be better but also Sergio has a point that the standard target isn't running tests (confirmed by checking the build log) so that isn't a functional change and should win us some build time/disk space.

@kenvandine seems reasonable?

kenvandine:
> I agree with Ken that running the tests would be better but also Sergio has a point that the standard target isn't running tests (confirmed by checking the build log) so that isn't a functional change and should win us some build time/disk space.
> 
> @kenvandine seems reasonable?

Yeah, sounds reasonable.

kenvandine (approved):
Looks good!

Issue #257, opened 2024-11-20
New candidate revision (r57) create icons issue


seb128:
Testing r57 with the evince candidate the icons in the decoration are missing, see screenshot

![evince](https://github.com/user-attachments/assets/9da5c95a-57c5-4c0f-b7e1-57afa5c11e65)

Also trying gnome-calculator from candidate

```
$ /snap/bin/gnome-calculator 
...
(gnome-calculator:22523): Gtk-WARNING **: 09:10:27.684: Failed to load icon /snap/gnome-calculator/965/data-dir/icons/Yaru/scalable/actions/edit-undo-symbolic.svg: Unable to load image-loading module: /snap/gnome-calculator/965/gnome-platform/usr/lib/x86_64-linux-gnu/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader-svg.so: /snap/gnome-calculator/965/gnome-platform/usr/lib/x86_64-linux-gnu/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader-svg.so: cannot open shared object file: No such file or directory
```

seb128:
In r48
/snap/gnome-46-2404/48/usr/lib/x86_64-linux-gnu/gdk-pixbuf-2.0/2.10.0/loaders.cache has 
"/build/gnome-46-2404-sdk/prime/usr/lib/x86_64-linux-gnu/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader-svg.so"

in r57
/snap/gnome-46-2404/57/usr/lib/x86_64-linux-gnu/gdk-pixbuf-2.0/2.10.0/loaders.cache
"/build/gnome-46-2404-sdk/prime/usr/lib/x86_64-linux-gnu/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader_svg.so"


libpixbufloader-svg.so change to libpixbufloader_svg.so for some reason?

sergio-costas:
I'll check it.

sergio-costas:
Mmm... odd... I'll try to rename that file.

sergio-costas:
Ok, it seems that snaps that use gnome extension do rebuild the gdk-pixbuf-loaders.cache at install or update. The problem is that, when the gnome-XX-XXXX.snap is updated, that cache isn't rebuilt, so any change would break that.

[...]

PR #263, opened 2024-11-25
Revert librsvg to version 2.58.1

sergio-costas:
This is required because version 2.59.1 changes the name of the gdk-pixbuf-loader from libpixbufloader-svg.so to
libpixbufloader_svg.so, which breaks snaps because each snap stores its own loaders.cache file with the path for each loader, so an already installed snap will have a loaders.cache file with a svg loader path pointing to the former name, but the current file would be the later, thus it won't be able to load any svg.

Reverting the version also forces to revert from meson to autotools.

The right solution is to modify snapcraft-desktop-integration scripts to add a check that rebuilds any cache (not only pixbuf loaders, but also fonts, icons...) not only if the snap itself is updated, but also if any of the extensions is updated too.

Fix https://github.com/ubuntu/gnome-sdk/issues/257

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -345,36 +345,27 @@ parts:
       - libtiff-dev
 
   librsvg:
-    after: [ gdk-pixbuf, vala, meson-deps ]
+    after: [ gdk-pixbuf, vala ]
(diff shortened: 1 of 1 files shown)


seb128 (approved):
Thanks!

PR #264, opened 2024-12-03
Update librsvg to version 2.59.1

sergio-costas:
This change requires using meson, and also to avoid snapcraft adding $CRAFT_STAGE to the `prefix` part in librsvg-2.0.pc, since now that .pc file is position-independent thanks to the use of ${pcfiledir}.

Also some changes to the libgweather patch had to be added to enable it to build.

Since the gdk-pixbuf loader is renamed, this PR should NOT be merged until https://github.com/canonical/snapcraft-desktop-integration/pull/30 is fully available in the build pipeline.

### DIFF
--- a/patches/libgweather.diff
+++ b/patches/libgweather.diff
@@ -6,7 +6,7 @@ index c4d6445a..9cec37cf 100644
    output: '@BASENAME@.bin',
    install: true,
    install_dir: pkglibdir,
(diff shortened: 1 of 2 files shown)


seb128 (approved):
The launcher changes landed so we can merge this one and give it a try in candidate now

seb128 (changes requested):
Oh, it needs manual rebasing first though, could you do that?

sergio-costas:
@seb128 Rebased and merged.

seb128:
Thanks!

PR #265, opened 2024-12-05
Better legibility for environment

sergio-costas:
The environment variables are very long. This PR cuts them in several lines to improve legibility and maintenability.


seb128 (approved):
thanks, it is easier to read indeed!

PR #266, opened 2024-12-05
Add HEIF/HEIC support

sergio-costas:
This patch adds headers and pixbuf-loader for HEIF/HEIC format.


seb128 (approved):
Thanks, it makes sense indeed

PR #270, opened 2024-12-12
Use .deb libtool instead of savannah version

seb128:
This is to avoid the downloading problems we are having from the Savannah git repo from GNU.


kenvandine (approved):
Looks good!

PR #271, opened 2025-01-16
libadwaita/patches: Add wartybrown accent color

3v1n0:
The Ubuntu 20th anniversary color has to be supported!

This patch is only needed until 22.10 EOL.

---

This patch was done long time ago, but apparently not subitted?!


kenvandine (approved):
Excellent, long live warty brown!

PR #272, opened 2025-01-16
libcanberra: Import patches from debian fixing wayland behavior

3v1n0:
Libcanberra has x11 checks that don't work well in wayland, so import the patches we've in distro already


sergio-costas (approved):
This PR looks fine.

sergio-costas:
I'm building it.

sergio-costas:
This builds fine, but can't build the "contents" snap from the sdk one due to a problem in my script. Anyway, it looks fine.

PR #274, opened 2025-01-23
[gnome-46] Compute version adding the sdk version as base

3v1n0:
The same snap extension can be built against different SDKs, so make it clearer in the revision.

This generates: `0+git.52c5c45-sdk0+git.b58afbc`

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,5 +1,5 @@
 name: gnome-46-2404
-version: git
+adopt-info: gnome-sdk
 summary: Shared GNOME 46 Ubuntu stack
 source-code: https://github.com/ubuntu/gnome-sdk/tree/gnome-46-2404
 description: |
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -36,6 +36,21 @@ parts:
   gnome-sdk:
     plugin: nil
     stage-snaps: [ gnome-46-2404-sdk/latest/candidate ]
+    build-packages:
+      - yq
+    override-build: |
+      set -eu
+      craftctl default
+      sdk_version=$(cat ${CRAFT_PART_INSTALL}/snap.gnome-46-2404-sdk/manifest.yaml \
```
sergio-costas, line 44: Wasn't sure if this would work, but checking the documentation saw that the "snap" folder is copied as `snap.snap-name`, so yes, this should work.

sergio-costas (approved):
LGTM.

PR #275, opened 2025-01-23
[gnome-42] Compute version adding the sdk version as base

3v1n0:
The same snap extension can be built against different SDKs, so make it clearer in the revision

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,5 +1,5 @@
 name: gnome-42-2204
-version: git
+adopt-info: gnome-sdk
 summary: Shared GNOME 42 Ubuntu stack
 source-code: https://github.com/ubuntu/gnome-sdk/tree/gnome-42-2204
 description: |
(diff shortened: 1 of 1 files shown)


3v1n0:
@sergio-costas this needs your love too :)

sergio-costas (approved):
It seems fine. Tested the "sed" filter for git version, and the result of git -C. But this last is "odd", in the sense that, since there aren't tags, it returns the git sha value... Shouldn't we use tags?

3v1n0:
> But this last is "odd", in the sense that, since there aren't tags, it returns the git sha value... Shouldn't we use tags?

Yeah, that's true, I think that's for discussion in another place, but ideally yes.

However this project doesn't define tags, so we can just rely on the sha. The code there is ready in case we start adding tags though.

sergio-costas:
Yeah, I agree in that it is something for another discussion. That's why I approved it anyway. Merging.

PR #277, opened 2025-02-13
Skip generated .debug files from extracting/stripping again

lissyx:


seb128 (approved):
Thanks, that makes sense. I'm a bit puzzled on why that wasn't an issue before, unless  `strip` used to not do anything when called on a .debug and is doing something now for some reason? Anyway the check makes sense

PR #281, opened 2025-03-20
Fix strip-binaries.py

sergio-costas:
In some libraries, the ELF module returns something that isn't a text string, but instead a bytes array, and trying to build a path with that breaks the execution.

This patch fixes this by translating the bytes into an hexadecimal string, as expected.

### DIFF
--- a/tools/strip_binaries.py
+++ b/tools/strip_binaries.py
@@ -41,6 +41,8 @@ def read_elf_type(filepath):
                 for note in sect.iter_notes():
                     if  note['n_type'] == 'NT_GNU_BUILD_ID':
                         buildId = note['n_desc']
+                        if isinstance(buildId, bytes):
+                            buildId = buildId.hex()
                         break
         except:
             return None, None


3v1n0:
What would be an example of invalid string?

sergio-costas:
For some reason, for libc_malloc_debug.so, it returns a bytes array instead of a string:
```
:: Error when extracting buildId from /root/prime/lib/x86_64-linux-gnu/libc_malloc_debug.so.0: b'\xa7&\x00\x00\x00\x00\x00\x00P\xb2\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00libc_malloc_debug\x00memory_mallopt_arena_max\x008@%rax 8@32+mp_(%rip)\x00'; retrying
```

sergio-costas:
The odd thing is that outside, it returns a true string...

```
./test_elf.py ../gnome-sdk/squashfs-root/lib/x86_64-linux-gnu/libc_malloc_debug.so.0 
('ET_DYN', 'a72600000000000050b200000000000000000000000000006c6962635f6d616c6c6f635f6465627567006d656d6f72795f6d616c6c6f70745f6172656e615f6d61780038402572617820384033322b6d705f28257269702900')
```

sergio-costas:
In fact, you can see the error here too: https://github.com/ubuntu/gnome-sdk/actions/runs/13975074520/job/39126695282

sergio-costas:
Ok, I rewrote the patch. Now it's safer.

nteodosio (approved):
The issue is indeed known: https://github.com/eliben/pyelftools/issues/188.

seb128:
Thanks for finding that upstream reference Nathan. @sergio-costas could you add a code comment with the upstream url for future reference? (feel free to directly push without going through a pull request if that's easier)

sergio-costas:
@seb128 Done. And also backported it to gnome-42-2404.

PR #283, opened 2025-03-21
Add reference to the issue fixed in the previous PR

sergio-costas:


nteodosio (approved):
Thanks for adding the comment!

PR #284, opened 2025-03-21
Backport the bugfix for debug symbols generator

sergio-costas:


nteodosio (approved):
Same as the previous merge request for 46-2404 so approving as well.

PR #288, opened 2025-04-03
Update Gtk4 and GtkMM

sergio-costas:
Gtk4 4.18.2 is quite buggy, so it must be updated to 4.18.3.

GtkMM 4.17.0 is a development version, it must be updated to 4.18.0. Also, ignore-odd-minor has been added to avoid this error in the future.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -551,7 +551,7 @@ parts:
   gtk4:
     after: [ wayland-protocols, harfbuzz, pango, meson-deps ]
     source: https://gitlab.gnome.org/GNOME/gtk.git
-    source-tag: '4.18.2'
+    source-tag: '4.18.3'
 # ext:updatesnap
 #   version-format:
 #     ignore-odd-minor: true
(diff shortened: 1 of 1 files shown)


sergio-costas:
@aleasto This should fix it.

aleasto (approved):
Thanks!

sergio-costas:
Tested with:

   * chromium: opening youtube and playing a video
   * gnome-recipes
   * shotwell
   * gnome-mahjongg
   * kicad-8
   * mattermost-desktop
   * evince
   * darktable
   * eog

All worked fine, although Darktable shows this:
```
(darktable:1310911): GLib-GIO-WARNING **: 11:47:30.804: Error creating IO channel for /proc/self/mountinfo: Permission denied (g-file-error-quark, 2)
     2.3028 [lib_load_module] failed to open `gamepad': libSDL2-2.0.so.0: cannot open shared object file: No such file or directory
```

seb128 (approved):
Let's do that before promoting a new build to stable!

aleasto:
>      2.3028 [lib_load_module] failed to open `gamepad': libSDL2-2.0.so.0: cannot open shared object file: No such file or directory

I guess nothing else in GNOME uses SDL2?

That would be a simple fix in darktable to bundle SDL2. The darktable documentation mentions that it only uses SDL2 for gamepad input. Although I wonder if that also requires poking a new little hole in the snap confinement

sergio-costas:
@aleasto My fault. Don't know why I presumed that gnome-sdk included SDL.

PR #289, opened 2025-04-04
Fix cargo-c build

sergio-costas:
Cargo dependencies seem to be weaker than required,and a recent change breaks building. This patch fixes it.

More info: https://github.com/lu-zero/cargo-c/pull/453


seb128 (approved):
Great, let's give that a try!

aleasto (approved):

sergio-costas:
It passed the point where it broke before. Let's hope that nothing else breaks when building librsvg...

PR #290, opened 2025-04-07
Update Gtk4 to 4.18.4

aleasto:
Another bugfix release, including the fix for LP: #2106299


seb128 (approved):
Thanks!

sergio-costas (approved):
Tested build locally, and it works. Also, the change is sensible.

PR #291, opened 2025-04-08
Fix dependencies for gobject-introspection and Vapigen

sergio-costas:
Some parts use gobject-introspection or vapigen, which require Vala and Gobject-introspection parts being built before. Unfortunately, some dependencies didn't account for this.

Also, the PATH environment was wrong, and didn't prioritize the STAGE, so in some cases the wrong binary was being used.

Due to all of this, building libportal failed when using gobject-introspection greater than 1.82.0.

This PR fixes all these things, and also adds an extra check in check_dependencies.py to ensure that any part that is building gobject-introspection or vapi files have the correct dependencies in the `after` entry. For this to work better, I also added missing meson options that builds gobject-introspection and vapi files (those options where already set to true by default, so they change nothing relative to the output snap; only ensures that dependencies will be correct).

### DIFF
--- a/.github/workflows/check-dependencies.yml
+++ b/.github/workflows/check-dependencies.yml
@@ -34,9 +34,9 @@ jobs:
           tools/check_dependencies.py test_data/test3.yaml
       - name: Invert success and failure for tests
         run: |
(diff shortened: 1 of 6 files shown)


seb128 (approved):
Thanks, seems to make sense!

PR #294, opened 2025-04-21
SDK: Fix so libraries dangling symlinks

3v1n0:
libraries .so files provided by -dev packages normally point to the
versioned so file, but if these files are provided by the core snap or
another base snap then we should instead use them, if they're not
provided by the SDK itself.

So, ensure that all the symlinks in the lib directory are actually
pointing to a library file, trying to resolve them with the
core-provided libraries as first try.

This fixes the compile issue we have with gnome-boxes as per the linking issues:

```
2025-04-21 20:49:32.887 :: /snap/gnome-46-2404-sdk/current/usr/bin/ld: /snap/gnome-46-2404-sdk/current/usr/lib/x86_64-linux-gnu/libpciaccess.a(linux_sysfs.o): warning: relocation against `stderr@@GLIBC_2.2.5' in read-only section `.text'
2025-04-21 20:49:32.887 :: /snap/gnome-46-2404-sdk/current/usr/bin/ld: /snap/gnome-46-2404-sdk/current/usr/lib/x86_64-linux-gnu/libpciaccess.a(common_device_name.o): in function `populate_vendor':
2025-04-21 20:49:32.887 :: (.text+0x18d): undefined reference to `gzopen'
2025-04-21 20:49:32.887 :: /snap/gnome-46-2404-sdk/current/usr/bin/ld: (.text+0x1bc): undefined reference to `gzgets'
2025-04-21 20:49:32.887 :: /snap/gnome-46-2404-sdk/current/usr/bin/ld: (.text+0x394): undefined reference to `gzclose'
2025-04-21 20:49:32.888 :: /snap/gnome-46-2404-sdk/current/usr/bin/ld: (.text+0x420): undefined reference to `gzopen'
2025-04-21 20:49:32.888 :: /snap/gnome-46-2404-sdk/current/usr/bin/ld: /snap/gnome-46-2404-sdk/current/usr/lib/x86_64-linux-gnu/libpciaccess.a(linux_sysfs.o): relocation R_X86_64_PC32 against symbol `stderr@@GLIBC_2.2.5' can not be used when making a shared object; recompile with -fPIC
```

This fixes also these other cases:

```
2025-04-21T19:14:27.7652675Z :: ++ readlink /root/prime/usr/lib/x86_64-linux-gnu/libBrokenLocale.so
2025-04-21T19:14:27.7655234Z :: + core_lib=/snap/core24/current/usr/lib/x86_64-linux-gnu/libBrokenLocale.so.1
2025-04-21T19:14:27.7657027Z :: + realpath --canonicalize-existing /snap/core24/current/usr/lib/x86_64-linux-gnu/libBrokenLocale.so.1
2025-04-21T19:14:27.7668303Z :: /snap/core24/888/usr/lib/x86_64-linux-gnu/libBrokenLocale.so.1
2025-04-21T19:14:27.7670954Z :: + ln -svf /usr/lib/x86_64-linux-gnu/libBrokenLocale.so.1 /root/prime/usr/lib/x86_64-linux-gnu/libBrokenLocale.so
2025-04-21T19:14:27.7682861Z :: '/root/prime/usr/lib/x86_64-linux-gnu/libBrokenLocale.so' -> '/usr/lib/x86_64-linux-gnu/libBrokenLocale.so.1'
2025-04-21T19:14:27.7685855Z :: + 
2025-04-21T19:14:27.8632229Z :: ++ readlink /root/prime/usr/lib/x86_64-linux-gnu/libanl.so
2025-04-21T19:14:27.8645632Z :: + core_lib=/snap/core24/current/usr/lib/x86_64-linux-gnu/libanl.so.1
2025-04-21T19:14:27.8648639Z :: + realpath --canonicalize-existing /snap/core24/current/usr/lib/x86_64-linux-gnu/libanl.so.1
2025-04-21T19:14:27.8658523Z :: /snap/core24/888/usr/lib/x86_64-linux-gnu/libanl.so.1
2025-04-21T19:14:27.8676057Z :: + ln -svf /usr/lib/x86_64-linux-gnu/libanl.so.1 /root/prime/usr/lib/x86_64-linux-gnu/libanl.so
2025-04-21T19:14:27.8676810Z :: '/root/prime/usr/lib/x86_64-linux-gnu/libanl.so' -> '/usr/lib/x86_64-linux-gnu/libanl.so.1'
2025-04-21T19:14:27.8679512Z :: + 
2025-04-21T19:14:27.9091535Z :: ++ readlink /root/prime/usr/lib/x86_64-linux-gnu/libc_malloc_debug.so
2025-04-21T19:14:27.9097459Z :: + core_lib=/snap/core24/current/usr/lib/x86_64-linux-gnu/libc_malloc_debug.so.0
2025-04-21T19:14:27.9098504Z :: + realpath --canonicalize-existing /snap/core24/current/usr/lib/x86_64-linux-gnu/libc_malloc_debug.so.0
2025-04-21T19:14:27.9107874Z :: /snap/core24/888/usr/lib/x86_64-linux-gnu/libc_malloc_debug.so.0
2025-04-21T19:14:27.9111709Z :: + ln -svf /usr/lib/x86_64-linux-gnu/libc_malloc_debug.so.0 /root/prime/usr/lib/x86_64-linux-gnu/libc_malloc_debug.so
2025-04-21T19:14:27.9126662Z :: '/root/prime/usr/lib/x86_64-linux-gnu/libc_malloc_debug.so' -> '/usr/lib/x86_64-linux-gnu/libc_malloc_debug.so.0'
2025-04-21T19:14:27.9128331Z :: + 
2025-04-21T19:14:27.9381751Z :: ++ readlink /root/prime/usr/lib/x86_64-linux-gnu/libcrypt.so
2025-04-21T19:14:27.9405472Z :: + core_lib=/snap/core24/current/usr/lib/x86_64-linux-gnu/libcrypt.so.1.1.0
2025-04-21T19:14:27.9406127Z :: + realpath --canonicalize-existing /snap/core24/current/usr/lib/x86_64-linux-gnu/libcrypt.so.1.1.0
2025-04-21T19:14:27.9407716Z :: /snap/core24/888/usr/lib/x86_64-linux-gnu/libcrypt.so.1.1.0
2025-04-21T19:14:27.9413478Z :: + ln -svf /usr/lib/x86_64-linux-gnu/libcrypt.so.1.1.0 /root/prime/usr/lib/x86_64-linux-gnu/libcrypt.so
2025-04-21T19:14:27.9427626Z :: '/root/prime/usr/lib/x86_64-linux-gnu/libcrypt.so' -> '/usr/lib/x86_64-linux-gnu/libcrypt.so.1.1.0'
2025-04-21T19:14:27.9428341Z :: + 
2025-04-21T19:14:28.2410903Z :: ++ readlink /root/prime/usr/lib/x86_64-linux-gnu/libmvec.so
2025-04-21T19:14:28.2424731Z :: + core_lib=/snap/core24/current/usr/lib/x86_64-linux-gnu/libmvec.so.1
2025-04-21T19:14:28.2425665Z :: + realpath --canonicalize-existing /snap/core24/current/usr/lib/x86_64-linux-gnu/libmvec.so.1
2025-04-21T19:14:28.2433988Z :: /snap/core24/888/usr/lib/x86_64-linux-gnu/libmvec.so.1
2025-04-21T19:14:28.2437987Z :: + ln -svf /usr/lib/x86_64-linux-gnu/libmvec.so.1 /root/prime/usr/lib/x86_64-linux-gnu/libmvec.so
2025-04-21T19:14:28.2448685Z :: '/root/prime/usr/lib/x86_64-linux-gnu/libmvec.so' -> '/usr/lib/x86_64-linux-gnu/libmvec.so.1'
2025-04-21T19:14:28.2451476Z :: + 
2025-04-21T19:14:28.2571158Z :: ++ readlink /root/prime/usr/lib/x86_64-linux-gnu/libnss_compat.so
2025-04-21T19:14:28.2575968Z :: + core_lib=/snap/core24/current/usr/lib/x86_64-linux-gnu/libnss_compat.so.2
2025-04-21T19:14:28.2576937Z :: + realpath --canonicalize-existing /snap/core24/current/usr/lib/x86_64-linux-gnu/libnss_compat.so.2
2025-04-21T19:14:28.2590023Z :: /snap/core24/888/usr/lib/x86_64-linux-gnu/libnss_compat.so.2
2025-04-21T19:14:28.2591451Z :: + ln -svf /usr/lib/x86_64-linux-gnu/libnss_compat.so.2 /root/prime/usr/lib/x86_64-linux-gnu/libnss_compat.so
2025-04-21T19:14:28.2603414Z :: '/root/prime/usr/lib/x86_64-linux-gnu/libnss_compat.so' -> '/usr/lib/x86_64-linux-gnu/libnss_compat.so.2'
2025-04-21T19:14:28.2617420Z :: + 
2025-04-21T19:14:28.2633341Z :: ++ readlink /root/prime/usr/lib/x86_64-linux-gnu/libnss_hesiod.so
2025-04-21T19:14:28.2638364Z :: + core_lib=/snap/core24/current/usr/lib/x86_64-linux-gnu/libnss_hesiod.so.2
2025-04-21T19:14:28.2647972Z :: + realpath --canonicalize-existing /snap/core24/current/usr/lib/x86_64-linux-gnu/libnss_hesiod.so.2
2025-04-21T19:14:28.2651783Z :: /snap/core24/888/usr/lib/x86_64-linux-gnu/libnss_hesiod.so.2
2025-04-21T19:14:28.2654605Z :: + ln -svf /usr/lib/x86_64-linux-gnu/libnss_hesiod.so.2 /root/prime/usr/lib/x86_64-linux-gnu/libnss_hesiod.so
2025-04-21T19:14:28.2672222Z :: '/root/prime/usr/lib/x86_64-linux-gnu/libnss_hesiod.so' -> '/usr/lib/x86_64-linux-gnu/libnss_hesiod.so.2'
2025-04-21T19:14:28.2674379Z :: + 
2025-04-21T19:14:28.2958314Z :: ++ readlink /root/prime/usr/lib/x86_64-linux-gnu/libpciaccess.so
2025-04-21T19:14:28.2972622Z :: + core_lib=/snap/core24/current/usr/lib/x86_64-linux-gnu/libpciaccess.so.0.11.1
2025-04-21T19:14:28.2973906Z :: + realpath --canonicalize-existing /snap/core24/current/usr/lib/x86_64-linux-gnu/libpciaccess.so.0.11.1
2025-04-21T19:14:28.2983570Z :: /snap/core24/888/usr/lib/x86_64-linux-gnu/libpciaccess.so.0.11.1
2025-04-21T19:14:28.2990439Z :: + ln -svf /usr/lib/x86_64-linux-gnu/libpciaccess.so.0.11.1 /root/prime/usr/lib/x86_64-linux-gnu/libpciaccess.so
2025-04-21T19:14:28.3002763Z :: '/root/prime/usr/lib/x86_64-linux-gnu/libpciaccess.so' -> '/usr/lib/x86_64-linux-gnu/libpciaccess.so.0.11.1'
2025-04-21T19:14:28.3006533Z :: + 
2025-04-21T19:14:28.3782845Z :: ++ readlink /root/prime/usr/lib/x86_64-linux-gnu/libresolv.so
2025-04-21T19:14:28.3797871Z :: + core_lib=/snap/core24/current/usr/lib/x86_64-linux-gnu/libresolv.so.2
2025-04-21T19:14:28.3799269Z :: + realpath --canonicalize-existing /snap/core24/current/usr/lib/x86_64-linux-gnu/libresolv.so.2
2025-04-21T19:14:28.3808636Z :: /snap/core24/888/usr/lib/x86_64-linux-gnu/libresolv.so.2
2025-04-21T19:14:28.3822038Z :: + ln -svf /usr/lib/x86_64-linux-gnu/libresolv.so.2 /root/prime/usr/lib/x86_64-linux-gnu/libresolv.so
2025-04-21T19:14:28.3826156Z :: '/root/prime/usr/lib/x86_64-linux-gnu/libresolv.so' -> '/usr/lib/x86_64-linux-gnu/libresolv.so.2'
2025-04-21T19:14:28.3827052Z :: + 
2025-04-21T19:14:28.4324023Z :: ++ readlink /root/prime/usr/lib/x86_64-linux-gnu/libthread_db.so
2025-04-21T19:14:28.4338047Z :: + core_lib=/snap/core24/current/usr/lib/x86_64-linux-gnu/libthread_db.so.1
2025-04-21T19:14:28.4339762Z :: + realpath --canonicalize-existing /snap/core24/current/usr/lib/x86_64-linux-gnu/libthread_db.so.1
2025-04-21T19:14:28.4351158Z :: /snap/core24/888/usr/lib/x86_64-linux-gnu/libthread_db.so.1
2025-04-21T19:14:28.4355710Z :: + ln -svf /usr/lib/x86_64-linux-gnu/libthread_db.so.1 /root/prime/usr/lib/x86_64-linux-gnu/libthread_db.so
2025-04-21T19:14:28.4368474Z :: '/root/prime/usr/lib/x86_64-linux-gnu/libthread_db.so' -> '/usr/lib/x86_64-linux-gnu/libthread_db.so.1'
2025-04-21T19:14:28.4381415Z :: + 
```

UDENG-6680

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1606,10 +1606,6 @@ parts:
       cd $CRAFT_STAGE/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR
       find . -type f,l -exec rm -fv $CRAFT_PART_INSTALL/usr/lib/{} \;
       find . -type f,l -name "*.so.*" -exec sh -c "rm -fv $CRAFT_PART_INSTALL/usr/lib/{}*" \;
-      # FIXME, need to figure out why libayatana-appindicator install those
-      rm -f $CRAFT_PART_INSTALL/usr/lib/*/libatk-1.0.so*
-      rm -f $CRAFT_PART_INSTALL/usr/lib/*/libatk-bridge-2.0.so*
-      rm -f $CRAFT_PART_INSTALL/usr/lib/*/libatspi.so*
     stage:
       - -usr/bin/fc-cache
 
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1748,11 +1748,6 @@ parts:
       rm -rf usr/share/locale-langpack/kab
       rm -f usr/share/unity/client-scopes.json
 
-      # Libm is a dangling symlink in armhf and riscv64, breaking builds.
-      # We don't need them anyway, the base image has them.
-      rm -f usr/lib/*/lib[mc].so || :
-      rm -f usr/lib/*/lib[mc].a  || :
-
       find . -type d -empty -delete
 
       PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1456,10 +1456,12 @@ parts:
       - libdrm2
       - libdrm-dev
       - libegl1-mesa-dev
+      - libelf1t64
       - libevdev2
       - libevdev-dev
       - libexpat1
       - libexpat1-dev
+      - libffi8
       - libfontconfig1-dev
       - libfreetype6
       - libfreetype-dev
@@ -1481,6 +1483,7 @@ parts:
       - libgraphene-1.0-dev
       - libgraphviz-dev
       - libgspell-1-dev
+      - libgssapi-krb5-2
       - libgstreamer-plugins-bad1.0-dev
       - libgstreamer-plugins-base1.0-dev
       - libgstreamer-plugins-good1.0-dev
@@ -1491,6 +1494,8 @@ parts:
       - libidn2-0
       - libidn2-dev
       - libjpeg-dev
+      - libk5crypto3
+      - libkrb5support0
       - libkrb5-3
       - libkrb5-dev
       - liblcms2-dev
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1701,7 +1701,7 @@ parts:
       if [ ${#invalid_links[@]} -gt 0 ]; then
         echo "Invalid symlinks found:"
         echo "${invalid_links[@]}"
-        echo "Ensure full package set is staged!"
+        echo "Ensure full package set (runtime and dev files) is staged!"
         exit 1;
       fi
 
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1682,10 +1682,27 @@ parts:
       ln -s gettext-0.19.8 usr/share/gettext-current
       ln -s gdk-pixbuf-2.0/2.10.0 usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/gdk-pixbuf-current
 
-      # Fix dangling symlink by overwriting it
-      ln -sf lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libc_malloc_debug.so.0 usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libc_malloc_debug.so
-      # CHECK THIS When changing the core base, this line must be update
-      cp /snap/$CRAFT_EXT_CORE_LEVEL/current/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libc_malloc_debug.so.0 lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/
+      # Check dangling libraries symlinks
+      invalid_links=()
+      core_root_prefix=/snap/$CRAFT_EXT_CORE_LEVEL/current
+      for i in $(find "$CRAFT_PRIME"/{usr/,}lib -xtype l -name '*.so'); do
+        # Try fix dangling symlink by overwriting it with core library if any.
+        rel_lib_path=$(realpath --relative-to="$CRAFT_PRIME" "$i")
+        core_lib=$core_root_prefix/$(dirname "$rel_lib_path")/$(readlink "$i")
+
+        if realpath --canonicalize-existing "$core_lib"; then
+          ln -svf "${core_lib#"$core_root_prefix"}" "$i"
+          continue
+        fi
+
+        echo "$i: dangling symlink!"
+        invalid_links+=($i)
+      done
+      if [ ${#invalid_links[@]} -gt 0 ]; then
+        echo "Invalid symlinks found:"
+        echo "${invalid_links[@]}"
+        exit 1;
+      fi
 
       # Necessary for armhf builds, triggers review warning
       if [ "$CRAFT_ARCH_TRIPLET_BUILD_FOR" = "arm-linux-gnueabihf" ]; then
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1443,6 +1443,8 @@ parts:
       - libbrotli-dev
       - libbz2-1.0
       - libbz2-dev
+      - libc6
+      - libc6-dev
       - libcairo2-dev
       - libcanberra-gtk3-dev
       - libcanberra-pulse
@@ -1506,6 +1508,8 @@ parts:
       - libnsl-dev
       - libnss3-dev
       - libopenjp2-7-dev
+      - libpciaccess0
+      - libpciaccess-dev
       - libpcre2-8-0
       - libpcre2-dev
       - libpcre3
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1592,15 +1592,18 @@ parts:
     override-build: |
       set -eux
       craftctl default
+
+      # Drop all the duplicates *.so.* library files that have been pulled from
+      # deb files and that we're already shipping as part of source-compiled parts
       cd $CRAFT_STAGE/usr
-      find . -type f,l -exec rm -f $CRAFT_PART_INSTALL/usr/{} \;
-      find . -type f,l -name "*.so*" -exec bash -c "rm -f $CRAFT_PART_INSTALL/usr/{}*" \;
+      find . -type f,l -exec rm -fv $CRAFT_PART_INSTALL/usr/{} \;
+      find . -type f,l -name "*.so.*" -exec sh -c "rm -fv $CRAFT_PART_INSTALL/usr/{}*" \;
       cd $CRAFT_STAGE/usr/lib
-      find . -type f,l -exec rm -f $CRAFT_PART_INSTALL/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/{} \;
-      find . -type f,l -name "*.so*" -exec bash -c "rm -f $CRAFT_PART_INSTALL/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/{}*" \;
+      find . -type f,l -exec rm -fv $CRAFT_PART_INSTALL/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/{} \;
+      find . -type f,l -name "*.so.*" -exec sh -c "rm -fv $CRAFT_PART_INSTALL/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/{}*" \;
       cd $CRAFT_STAGE/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR
-      find . -type f,l -exec rm -f $CRAFT_PART_INSTALL/usr/lib/{} \;
-      find . -type f,l -name "*.so*" -exec bash -c "rm -f $CRAFT_PART_INSTALL/usr/lib/{}*" \;
+      find . -type f,l -exec rm -fv $CRAFT_PART_INSTALL/usr/lib/{} \;
+      find . -type f,l -name "*.so.*" -exec sh -c "rm -fv $CRAFT_PART_INSTALL/usr/lib/{}*" \;
       # FIXME, need to figure out why libayatana-appindicator install those
       rm -f $CRAFT_PART_INSTALL/usr/lib/*/libatk-1.0.so*
       rm -f $CRAFT_PART_INSTALL/usr/lib/*/libatk-bridge-2.0.so*
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1691,21 +1691,13 @@ parts:
       invalid_links=()
       core_root_prefix=/snap/$CRAFT_EXT_CORE_LEVEL/current
       for i in $(find "$CRAFT_PRIME"/{usr/,}lib -xtype l -name '*.so'); do
-        # Try fix dangling symlink by overwriting it with core library if any.
-        rel_lib_path=$(realpath --relative-to="$CRAFT_PRIME" "$i")
-        core_lib=$core_root_prefix/$(dirname "$rel_lib_path")/$(readlink "$i")
-
-        if realpath --canonicalize-existing "$core_lib"; then
-          ln -svf "${core_lib#"$core_root_prefix"}" "$i"
-          continue
-        fi
-
         echo "$i: dangling symlink!"
         invalid_links+=($i)
       done
       if [ ${#invalid_links[@]} -gt 0 ]; then
         echo "Invalid symlinks found:"
         echo "${invalid_links[@]}"
+        echo "Ensure full package set is staged!"
         exit 1;
       fi
 
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1448,7 +1448,9 @@ parts:
       - libcairo2-dev
       - libcanberra-gtk3-dev
       - libcanberra-pulse
+      - libcrypt1
       - libcrypt-dev
+      - libcom-err2
       - libcurl4-openssl-dev
       - libdbus-1-3
       - libdbus-1-dev
@@ -1546,6 +1548,7 @@ parts:
       - libtasn1-6-dev
       - libtdb-dev
       - libthai-dev
+      - libtirpc3t64
       - libudev1
       - libudev-dev
       - libunity-dev
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -90,8 +90,6 @@ parts:
     override-stage: |
       set -eux
       craftctl default
-      rm $CRAFT_STAGE/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/lib[mc].a || :
-      rm $CRAFT_STAGE/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/lib[mc].so || :
       LIBTOOLIZE=usr/bin/libtoolize
       sed -i 's#pkgauxdir="#pkgauxdir="$CRAFT_STAGE#' $LIBTOOLIZE
       sed -i 's#pkgltdldir="#pkgltdldir="$CRAFT_STAGE#' $LIBTOOLIZE


snapcraft.yaml
```diff
@@ -1626,10 +1637,29 @@ parts:
...
-      # Fix dangling symlink by overwriting it
-      ln -sf lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libc_malloc_debug.so.0 usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libc_malloc_debug.so
-      # CHECK THIS When changing the core base, this line must be update
-      cp /snap/$CRAFT_EXT_CORE_LEVEL/current/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libc_malloc_debug.so.0 lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/
+      # Check dangling libraries symlinks
+      invalid_links=()
+      core_root_prefix=/snap/$CRAFT_EXT_CORE_LEVEL/current
+      for i in "$CRAFT_PRIME"/{,usr}/lib/{"$CRAFT_ARCH_TRIPLET_BUILD_FOR",}/*.so; do
+        if [ -L "$i" ] && ! realpath --canonicalize-existing "$i"; then
+          # Try fix dangling symlink by overwriting it with core library if any.
```
sergio-costas: Is it really needed to point them into the corresponding library in Core? Shouldn't it be enough to delete them, and let the search in LD_LIBRARY_PATH find them?

sergio-costas (commented):
A little nitpick

snapcraft.yaml
3v1n0: I wanted to mention it, since this was my idea too initially, but... Sadly no.

This is because the `.so` files are part of the `-dev` packages and as per this they are not part of the core, thus we need to ensure they are in the sdk, and given that these are not overriden...

So in short we need the so to be somewhere and to be linked to a valid path.

sergio-costas: Ok, so I'm doing some tests, and I see, for example, that libthread_db.so does exist only in gnome-46-2404-sdk, but not in core24, and points to libthread_db.so.1; but libthread_db.so.1 doesn't exist in gnome-46-2404-sdk, but in core24...

This looks more like a bug in core24... those links should exist there too, don't you think?

3v1n0: > Ok, so I'm doing some tests, and I see, for example, that libthread_db.so does exist only in gnome-46-2404-sdk, but not in core24, and points to libthread_db.so.1; but libthread_db.so.1 doesn't exist in gnome-46-2404-sdk, but in core24
> This looks more like a bug in core24... those links should exist there too, don't you think?

Not sure I understood it fully, but in core24 I have this

```
marco@tricky:/var/lib/snapd/void$ cat /etc/os-release
NAME="Ubuntu Core"
VERSION="24"
ID=ubuntu-core
PRETTY_NAME="Ubuntu Core 24"
VERSION_ID="24"
HOME_URL="https://snapcraft.io/"
BUG_REPORT_URL="https://bugs.launchpad.net/snappy/"
marco@tricky:/var/lib/snapd/void$ ls -l /usr/lib/x86_64-linux-gnu/libthread_db.so.1
-rw-r--r-- 1 root root 47912 Aug  8  2024 /usr/lib/x86_64-linux-gnu/libthread_db.so.1
marco@tricky:/var/lib/snapd/void$ ls -l /usr/lib/x86_64-linux-gnu/libthread_db.so
ls: cannot access '/usr/lib/x86_64-linux-gnu/libthread_db.so': No such file or directory
```

Now, that's fine because it's a *runtime* snap, so there's no need for the `*.so` file since the application will link with the versioned library instead (`libthread_db.so.1`).

When building instead, in order to be able to know to what version we're linking to, the linker will just use `libthread_db.so` that eventually will be resolved to the versioned library for the loader.

As per this, in the SDK we do need to have correctly linked `.so` files, while in the runtime these are unneeded (in fact they will be removed when building the non-sdk snap).

sergio-costas: Of course!!! You are right, sorry.

sergio-costas: What's odd is that it allows to do a soft link outside the snap... I presumed that it wasn't allowed by snapcraft...

sergio-costas (approved):
It looks fine, and even has a check to notify for any left dangling symlink.

sergio-costas:
Before merging this... what happens with the review tools? It seems that they do ensure that no soft link points outside the snap...

sergio-costas:
So maybe the solution is to add all those links in Core24...

3v1n0:
> So maybe the solution is to add all those links in Core24...

Oh I didn't check review tools output... But if they don't work with this, then... The only option would be to copy the actual libraries in the sdk too and ensure that they don't get removed as duplicate...

sergio-costas:
> Oh I didn't check review tools output... But if they don't work with this, then... The only option would be to copy the actual libraries in the sdk too and ensure that they don't get removed as duplicate...

I still think that they should go in Core24: if you think about it, it IS being used to compile, when you build any snap...

sergio-costas:
I really dislike having duplicated libraries.

3v1n0:
> I really dislike having duplicated libraries.

Yeah, but I feel that's the way... Since I'm pretty sure that core doesn't want to include dev packages contents.

While us being a dev sdk we need to provide everything required to build.

So if core is missing something, we have to.

Implies more maintenance though, I agree.

sergio-costas:
So... do we copy the required files from CoreXX into the right folder wherever there is a dangling pointer...?

3v1n0:
AFAIK is the only way we can... And ensure they get not deleted by duplicates checking tool

snapcraft.yaml
```diff
@@ -1626,10 +1637,29 @@ parts:
...
       ln -s gdk-pixbuf-2.0/2.10.0 usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/gdk-pixbuf-current
 
-      # Fix dangling symlink by overwriting it
-      ln -sf lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libc_malloc_debug.so.0 usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libc_malloc_debug.so
-      # CHECK THIS When changing the core base, this line must be update
-      cp /snap/$CRAFT_EXT_CORE_LEVEL/current/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libc_malloc_debug.so.0 lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/
+      # Check dangling libraries symlinks
+      invalid_links=()
+      core_root_prefix=/snap/$CRAFT_EXT_CORE_LEVEL/current
+      for i in "$CRAFT_PRIME"/{,usr}/lib/{"$CRAFT_ARCH_TRIPLET_BUILD_FOR",}/*.so; do
```
nteodosio: Wouldn't we be relying on luck that neither the SDK nor the core snaps provide so files in deeper levels not reached by these globs? For that, `find "$CRAFT_PRIME"/{usr/,}lib -xtype l -name '*.so'` would be more robust.

nteodosio (changes requested):

snapcraft.yaml
sergio-costas: If that's the case, then I recommend a python script. It can be more clever... (like I'm doing with https://github.com/sergio-costas/snap-build-tools.git , remove_common.py)

3v1n0: Yeah, I guess remove-common has to be adjusted at this point not to remove the duplicates where a *.so file points to them in the sdk, right?

nteodosio:
Moving from #296: 

> The point is that if the SDK requires something in the CoreXX to be built, it must add the corresponding -dev package in `build-packages` too, not build itself from scratch, and those soft links will not be in `stage` but in the build system. In `stage` will be only pieces installed due to dependencies, and those must be removed.

What do you mean with "build itself from scratch", what is "itself"? In Libtool's case the -dev package (libc6-dev) was added to build-packages.

sergio-costas:
> What do you mean with "build itself from scratch", what is "itself"? In Libtool's case the -dev package (libc6-dev) was added to build-packages.

Exactly: since Core24 doesn't have the soft link `libBrokenLocale.so` pointing to `libBrokenLocale.so.1`, then the SDK snap has to install libc6-dev, which is the one that creates it, instead of building libc6 itself from sources to have that link.

snapcraft.yaml
sergio-costas: Not necessarily. `remove-common` must be executed before any part has been built, but after all the `stage-packages` have been installed. This way, only duplicated files installed from .deb packages will be removed, but not an apparent duplicated file that was built in any part (thus we can build a new version of a library already being distributed in CoreXX or gnome-XX-YYYY-sdk). The *.so files must be removed in that case too, because those have been installed in STAGE by the stage-package list, so they aren't needed to build any of the packages. The ones that are needed to build are those installed in `build-packages`, which aren't installed in STAGE but in the building container itself.

sergio-costas:
In fact... why is libpciaccess in gnome-46-2404-sdk if it is already in Core24?

sergio-costas:
Ok, confirmed my suspicions: just by removing the broken soft link, the `libpciaccess.a` file, and (this is the key point) the `pciaccess.pc` file, all from the `gnome-46-2404-sdk'  snap, building gnome-boxes works as expected.

The problem is that the `pciaccess.pc` file in the `gnome-46-2404-sdk` snap takes precedence over the one installed from the `build-packages`, and they try to compile against a library in the snap. We removed the duplicated libraries (because they are already in Core24), but didn't remove the .pc file, and that is what is breaking everything.

So if we remove a library in the gnome SDK because it is already in Core24, we must also remove the corresponding .pc file to avoid this.

nteodosio:
> Exactly: since Core24 doesn't have the soft link libBrokenLocale.so pointing to libBrokenLocale.so.1, then the SDK snap has to install libc6-dev, which is the one that creates it, instead of building libc6 itself from sources to have that link.

That makes sense. We never built libc6 though.

sergio-costas:
@3v1n0 So, in my opinion, the true solution is to find which .pc file correspond to each duplicated library in Core24 and remove it. It's not easy, I now, but with some python magic... ;-)

sergio-costas:
In fact... just checking the "Libs" and "Libs.private" parts should be enough to know which libraries are affected by a .pc file, so if they aren't available in gnome-XX-YY-sdk, then that .pc file can be removed.

3v1n0:
> @3v1n0 So, in my opinion, the true solution is to find which .pc file correspond to each duplicated library in Core24 and remove it. It's not easy, I now, but with some python magic... ;-)

Well, I'm unsure either... These are build dependencies of dependencies we have in the SDK IIRC, I mean if any of these libraries headers/pc files are part of the gnome-sdk, then we should provide them too.

For example for the pciaccess one:

```
❯ apt-cache rdepends libpciaccess-dev
libpciaccess-dev
Reverse Depends:
  libdrm-dev
  xserver-xorg-dev
  xserver-xorg-dev
  xserver-xorg-dev
  libdrm-dev
  xserver-xorg-dev
  xserver-xorg-dev
  libdrm-dev
  xserver-xorg-dev
```

Since we have stage packages such as `libdrm-dev`, then we need to also fully stage their dependencies. And we were kinda doing it, but then we end up removing the duplicates that are part of core wrongly.

It's fine to remove them from the runtime snap, but not from the sdk.

sergio-costas:
But then, what we have to do is to ensure that using the gnome SDK requires to install the corresponding stage-packages that are provided by CoreXX.

sergio-costas:
Anyway... you don't necessarily need those dependencies to build something with those libraries... and if you need them for whatever reason, the build error will tell you what you have to add to the `build-packages`. And removing the pkgconfig files in the SDK will make that work.

sergio-costas:
Ok, let's take a step back and see the whole thing... Core24 (and other Cores) are built just by "dumping" an Ubuntu base system, as seen here: https://github.com/canonical/core-base/blob/277049e8bf67fdeadb9afe92428220e02cf95142/snapcraft.yaml#L13

The point is that, when building new snaps against CoreXX, the development files aren't there. Where are they get? And the answer is that `snapcraft` must automagically download them from .deb packages, things like `libc6-dev`, `build-essential`, `make`, `pkg-config`...

Also, there is a list of packages that must not be staged, because they are known to be included from Core... https://github.com/canonical/snapcraft/blob/0157e5b15e66af34a4b12d397e5f56a30eebb13d/snapcraft_legacy/internal/repo/_deb.py#L51

But all that I found at this point is only available in `legacy`, so for new snaps it must be done somewhere else.

So the point, in my opinion, is to find where those required -dev files are defined, and send a patch for snapcraft to add `libpciaccess0` and `libcrypt1` to the list of exceptions, and to include `libpciaccess-dev` and `libcrypt-dev`, because those two are already in Core24.

sergio-costas:
Ok, it seems that the list of blocked packages is in craft-parts: https://github.com/canonical/craft-parts/blob/1f163c0ec0691182cf08da52cc1cfc7160618241/craft_parts/packages/deb.py#L57-L283 So... Should we add there `libpciaccess0` and `libcrytp1`?

sergio-costas:
Also, in matrix https://matrix.to/#/!GGqzbFAUQprdPgYYCM:ubuntu.com/$bfuUc0FUzOVXuLGDn6RnlWXLWYqYzAVwfWdhF-qZRYE So it seems that no deb file is installed by snapcraft; instead, any development package seems to have to be installed by the snap, which is consistent with gnome-46-2404-sdk installing gcc, pkg-config, g++, clang...

sergio-costas:
Interesting thing: that list of packages in craft-parts is used for Core20 and previous; for others, it takes the list of packages installed in CoreXX and filters them https://github.com/canonical/craft-parts/blob/44aa6508874e8f4d022f13bd5760776ef987a79e/craft_parts/packages/deb.py#L362

Which means that libpciaccess0 will never be staged because it's already in Core24...

3v1n0:
As discussed in previous days, we should provide all the build-dependencies of our build-dependencies in the SDK, given that there's no way to symlink core paths we've to duplicate these core files in the SDK (for the runtime, I'll propose another patch to filter duplicates out)

The list of newly added files to the SDK is limited to:

```
Only in squashfs-root/usr/lib/x86_64-linux-gnu: gconv
Only in squashfs-root/usr/lib/x86_64-linux-gnu: ld-linux-x86-64.so.2
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libanl.so.1
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libBrokenLocale.so.1
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libc_malloc_debug.so.0
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libcrypt.so.1
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libcrypt.so.1.1.0
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libc.so.6
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libdl.so.2
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libmemusage.so
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libm.so.6
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libmvec.so.1
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libnsl.so.1
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libnss_compat.so.2
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libnss_dns.so.2
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libnss_files.so.2
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libnss_hesiod.so.2
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libpciaccess.so.0
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libpciaccess.so.0.11.1
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libpcprofile.so
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libpthread.so.0
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libresolv.so.2
Only in squashfs-root/usr/lib/x86_64-linux-gnu: librt.so.1
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libthread_db.so.1
Only in squashfs-root/usr/lib/x86_64-linux-gnu: libutil.so.1
Only in squashfs-root/usr: lib64
```

But at least now we have all the dev files without dead links.

Not sure if using bindings could have helped to avoid duplicates on core, but that would imply having a manually managed list of libraries to maintain.

snapcraft.yaml
```diff
@@ -1544,21 +1554,24 @@ parts:
       set -eux
       craftctl default
```
nteodosio, line 1604: Can we add a comment here explaining what we're undertaking in this override-build?

snapcraft.yaml
```diff
@@ -1544,21 +1554,24 @@ parts:
...
       cd $CRAFT_STAGE/usr
-      find . -type f,l -exec rm -f $CRAFT_PART_INSTALL/usr/{} \;
-      find . -type f,l -name "*.so*" -exec bash -c "rm -f $CRAFT_PART_INSTALL/usr/{}*" \;
+      find . -type f,l -exec rm -fv $CRAFT_PART_INSTALL/usr/{} \;
+      find . -type f,l -name "*.so.*" -exec sh -c "rm -fv $CRAFT_PART_INSTALL/usr/{}*" \;
       cd $CRAFT_STAGE/usr/lib
```
nteodosio, line 1611: Given that the Find command is recursive, why are we repeating it for `$CRAFT_STAGE/usr/lib` and `$CRAFT_STAGE/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR`?

snapcraft.yaml
```diff
@@ -1384,7 +1382,15 @@ parts:
       done
 
   debs:
-    after: [ libgnome-games-support, libadwaita, libgweather, libayatana-appindicator, libsoup3, librest, at-spi2-core, webp-pixbuf-loader ]
+    after:
```
sergio-costas, line 1403: these changes should be in a different MR, I think.

snapcraft.yaml
```diff
@@ -1628,10 +1641,19 @@ parts:
...
       ln -s gdk-pixbuf-2.0/2.10.0 usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/gdk-pixbuf-current
 
-      # Fix dangling symlink by overwriting it
-      ln -sf lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libc_malloc_debug.so.0 usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libc_malloc_debug.so
-      # CHECK THIS When changing the core base, this line must be update
-      cp /snap/$CRAFT_EXT_CORE_LEVEL/current/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libc_malloc_debug.so.0 lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/
+      # Check dangling libraries symlinks
+      invalid_links=()
+      core_root_prefix=/snap/$CRAFT_EXT_CORE_LEVEL/current
+      for i in $(find "$CRAFT_PRIME"/{usr/,}lib/{"$CRAFT_ARCH_TRIPLET_BUILD_FOR",} -xtype l -name '*.so'); do
```
nteodosio: Given that the Find command is recursive, this is equivalent to `$CRAFT_PRIME/{usr/,}lib`.

snapcraft.yaml
```diff
@@ -1628,10 +1641,19 @@ parts:
       ln -s gettext-0.19.8 usr/share/gettext-current
       ln -s gdk-pixbuf-2.0/2.10.0 usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/gdk-pixbuf-current
 
-      # Fix dangling symlink by overwriting it
-      ln -sf lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libc_malloc_debug.so.0 usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libc_malloc_debug.so
-      # CHECK THIS When changing the core base, this line must be update
-      cp /snap/$CRAFT_EXT_CORE_LEVEL/current/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libc_malloc_debug.so.0 lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/
+      # Check dangling libraries symlinks
```
sergio-costas, line 1694: I like this.

snapcraft.yaml
```diff
@@ -1544,21 +1554,24 @@ parts:
       set -eux
       craftctl default
       cd $CRAFT_STAGE/usr
-      find . -type f,l -exec rm -f $CRAFT_PART_INSTALL/usr/{} \;
-      find . -type f,l -name "*.so*" -exec bash -c "rm -f $CRAFT_PART_INSTALL/usr/{}*" \;
+      find . -type f,l -exec rm -fv $CRAFT_PART_INSTALL/usr/{} \;
+      find . -type f,l -name "*.so.*" -exec sh -c "rm -fv $CRAFT_PART_INSTALL/usr/{}*" \;
```
nteodosio, line 1610: `"*.so*"` -> `"*.so.*"`: Are we intentionally dropping files ending in `.so` here? Why?

nteodosio (changes requested):

snapcraft.yaml
```diff
@@ -1401,6 +1407,8 @@ parts:
       - libblkid1
       - libbrotli1
       - libbrotli-dev
+      - libc6
+      - libc6-dev
       - libcairo2-dev
```
sergio-costas, line 1448: (not related with the patch) Mmm... if we are already compiling cairo, why do we add this...?

snapcraft.yaml
```diff
@@ -1463,6 +1471,8 @@ parts:
       - libnss3-dev
       - libopenjp2-7
       - libopenjp2-7-dev
+      - libpciaccess0
```
sergio-costas, line 1518: If we are doing this, then we also should add libcrypt-dev and libcrypt1. They are missing too.

sergio-costas (changes requested):
Some details.

snapcraft.yaml
3v1n0: Oh, no idea... Feel free to open another MR for this. It's likely du to some old behavior

snapcraft.yaml
3v1n0: Do we have any dev r-dependency on it? Since I didn't see it being staged anywhere (at least we have no *.so file for it in the snap)

snapcraft.yaml
3v1n0: We were dropping them, now we don't do it anymore.

We only drop the `*.so.*` files now, that includes all the so-versions for a library we already ship from compiled sources

snapcraft.yaml
3v1n0: Oh, right... Indeed!

snapcraft.yaml
sergio-costas:
```
raster /snap/core24/current $ find |grep libcrypt
./usr/lib/x86_64-linux-gnu/cryptsetup/libcryptsetup-token-systemd-fido2.so
./usr/lib/x86_64-linux-gnu/cryptsetup/libcryptsetup-token-systemd-pkcs11.so
./usr/lib/x86_64-linux-gnu/cryptsetup/libcryptsetup-token-systemd-tpm2.so
./usr/lib/x86_64-linux-gnu/libcrypt.so.1
./usr/lib/x86_64-linux-gnu/libcrypt.so.1.1.0
./usr/lib/x86_64-linux-gnu/libcrypto.so.3
./usr/lib/x86_64-linux-gnu/libcryptsetup.so.12
./usr/lib/x86_64-linux-gnu/libcryptsetup.so.12.10.0
./usr/share/doc/libcrypt1
./usr/share/doc/libcrypt1/copyright
./usr/share/doc/libcryptsetup12
./usr/share/doc/libcryptsetup12/copyright
```

It's there...

sergio-costas: and the SDK contains it...

```
raster /snap/gnome-46-2404-sdk/current $ find |grep libcrypt
./usr/lib/x86_64-linux-gnu/libcrypt.a
./usr/lib/x86_64-linux-gnu/libcrypt.so
./usr/lib/x86_64-linux-gnu/pkgconfig/libcrypt.pc
```

sergio-costas:
```
raster /snap/gnome-46-2404-sdk/current $ ls -l usr/lib/x86_64-linux-gnu/libcrypt.*
-rw-r--r-- 1 root root 271404 abr  8  2024 usr/lib/x86_64-linux-gnu/libcrypt.a
lrwxrwxrwx 1 root root     17 abr  8  2024 usr/lib/x86_64-linux-gnu/libcrypt.so -> libcrypt.so.1.1.0
```

sergio-costas: Anyway, we must either add libcrypt1 and libcrypt-dev, or remove libcrypt.pc...

3v1n0: Ah, ok... Indeed is there, but as the script checks already in the snap we build from this branch we get:

```
squashfs-root/usr/lib/x86_64-linux-gnu/libcrypt.a
squashfs-root/usr/lib/x86_64-linux-gnu/libcrypt.so
squashfs-root/usr/lib/x86_64-linux-gnu/libcrypt.so.1
squashfs-root/usr/lib/x86_64-linux-gnu/libcrypt.so.1.1.0
```

So everything is added there already (if the symlink was wrong we were failing already), so if you want we can make the dependency explicit, but everything is already properly pulled in

sergio-costas: Oh, so this patch does fix that too... perfect then.

snapcraft.yaml
3v1n0: It wasn't my code, but I think the reason was to ensure that libfoo.so.X is both removed from `/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR` and `/usr/lib`, in case it got installed in different paths.

However I agree this can be probably be improved, but I think it's more material for another branch, so feel free to open a PR once this merges.

3v1n0: @sergio-costas may have further hints

nteodosio (approved):

snapcraft.yaml
sergio-costas: No idea, it was that way when I arrived :-D

Now seriously, this is already in `gnome-3-38-2004-sdk`, so it seems something that "slipped". Probably whoever added it just said "libraries are here and here, so I'll put a command for each folder", without realizing that `find` is recursive.

snapcraft.yaml
```diff
@@ -1543,19 +1545,18 @@ parts:
...
+      find . -type f,l -name "*.so.*" -exec sh -c "rm -fv $CRAFT_PART_INSTALL/usr/{}*" \;
       cd $CRAFT_STAGE/usr/lib
-      find . -type f,l -exec rm -f $CRAFT_PART_INSTALL/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/{} \;
-      find . -type f,l -name "*.so*" -exec bash -c "rm -f $CRAFT_PART_INSTALL/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/{}*" \;
+      find . -type f,l -exec rm -fv $CRAFT_PART_INSTALL/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/{} \;
+      find . -type f,l -name "*.so.*" -exec sh -c "rm -fv $CRAFT_PART_INSTALL/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/{}*" \;
       cd $CRAFT_STAGE/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR
-      find . -type f,l -exec rm -f $CRAFT_PART_INSTALL/usr/lib/{} \;
-      find . -type f,l -name "*.so*" -exec bash -c "rm -f $CRAFT_PART_INSTALL/usr/lib/{}*" \;
-      # FIXME, need to figure out why libayatana-appindicator install those
```
sergio-costas, line 1604: Mmmm... is this already fixed?

sergio-costas (commented):
Just a detail.

snapcraft.yaml
3v1n0: Yeah, and we only stage these files:

```
      - usr/lib/*/*indica*
      - usr/lib/*/*dbusmenu*
```

sergio-costas (approved):
LGTM.

sergio-costas:
It seems that krb5-multidev is either required, o removing the dangling pointer `/usr/lib/x86_64-linux-gnu/mit-krb5/libgssapi_krb5.so` is needed...

3v1n0:
> It seems that krb5-multidev is either required, o removing the dangling pointer `/usr/lib/x86_64-linux-gnu/mit-krb5/libgssapi_krb5.so` is needed...

It looks a bit out of place here TBH. @nteodosio do you know what could have pulled it in?

nteodosio:
Yes, libkrb5-dev, which in turn is pulled by Soup3.

sergio-costas:
@nteodosio It fails to build...

nteodosio:
Thanks, with the last commit we went from
```
/root/prime/usr/lib/x86_64-linux-gnu/libcom_err.so: dangling symlink!
/root/prime/usr/lib/x86_64-linux-gnu/libcrypt.so: dangling symlink!
/root/prime/usr/lib/x86_64-linux-gnu/libelf.so: dangling symlink!
/root/prime/usr/lib/x86_64-linux-gnu/libffi.so: dangling symlink!
/root/prime/usr/lib/x86_64-linux-gnu/libgssapi_krb5.so: dangling symlink!
/root/prime/usr/lib/x86_64-linux-gnu/libk5crypto.so: dangling symlink!
/root/prime/usr/lib/x86_64-linux-gnu/libkrb5support.so: dangling symlink!
/root/prime/usr/lib/x86_64-linux-gnu/libtirpc.so: dangling symlink!
/root/prime/usr/lib/x86_64-linux-gnu/mit-krb5/libgssapi_krb5.so: dangling symlink!
/root/prime/usr/lib/x86_64-linux-gnu/mit-krb5/libk5crypto.so: dangling symlink!
/root/prime/usr/lib/x86_64-linux-gnu/mit-krb5/libkrb5support.so: dangling symlink!
```

down to
```
/root/prime/usr/lib/x86_64-linux-gnu/libk5crypto.so: dangling symlink!
/root/prime/usr/lib/x86_64-linux-gnu/libffi.so: dangling symlink!
/root/prime/usr/lib/x86_64-linux-gnu/libelf.so: dangling symlink!
/root/prime/usr/lib/x86_64-linux-gnu/libkrb5support.so: dangling symlink!
```

So still needing libffi8, libelf1t64, libk5crypto3 and libkrb5support0.

nteodosio:
Those are now gone but eventually more popped up.

```
:: /root/prime/usr/lib/x86_64-linux-gnu/libcrypt.so: dangling symlink!
:: /root/prime/usr/lib/x86_64-linux-gnu/libcom_err.so: dangling symlink!
:: /root/prime/usr/lib/x86_64-linux-gnu/libtirpc.so: dangling symlink!
```

Probably something needs to be re-worked, we can't go whack-a-moling ad infinitum.

3v1n0:
> Probably something needs to be re-worked, we can't go whack-a-moling ad infinitum.

Well, pkg-config can help us with that... We should just follow the whole dependency chain.

nteodosio:
All green, can we merge now?

sergio-costas:
> All green, can we merge now?

Did you build the RUNTIME from this, and tested it to ensure that everything works as expected?

nteodosio:
No, but I can. Is this requirement and the corresponding process documented somewhere?

sergio-costas:
It should be a requirement :-D I use lxcraft to build it, with these changes in the snapcraft.yaml:

```
@@ -35,14 +35,19 @@ slots:
 parts:
   gnome-sdk:
     plugin: nil
-    stage-snaps: [ gnome-46-2404-sdk/latest/candidate ]
+#    stage-snaps: [ gnome-46-2404-sdk/latest/candidate ]
     build-packages:
       - yq
     override-build: |
       set -eu
       craftctl default
-      sdk_version=$(cat ${CRAFT_PART_INSTALL}/snap.gnome-46-2404-sdk/manifest.yaml \
-        | yq -r '.version')
+      cp -a /snap/gnome-46-2404-sdk/current/usr $CRAFT_PART_INSTALL/
+      cp -a /snap/gnome-46-2404-sdk/current/etc $CRAFT_PART_INSTALL/
+      cp -a /snap/gnome-46-2404-sdk/current/var $CRAFT_PART_INSTALL/
+      cp -a /snap/gnome-46-2404-sdk/current/lib $CRAFT_PART_INSTALL/
+      #sdk_version=$(cat ${CRAFT_PART_INSTALL}/snap.gnome-46-2404-sdk/manifest.yaml \
+      #  | yq -r '.version')
+      sdk_version=46
 
       # Use the same logic of snapcraft
       project_version=$(git -C "${CRAFT_PROJECT_DIR}" describe --dirty 2>/dev/null || true)
```

and this _lxcraft.yaml_ configuration file:

```
image: ubuntu:noble

force_debug: True

snaps:
  snapcraft:
    - classic
    - stable
    - store
  core24:
    - store
  gtk-common-themes:
    - store
  ../gnome-sdk/gnome-46*.snap:
    - local
```
(of course, I built the SDK snap at `../gnome-sdk` folder

sergio-costas:
Here you can see the list of snaps that I use to test each new SDK: https://github.com/ubuntu/gnome-sdk/pull/315

nteodosio:
Sorry, my connection is not only slow but today also flaky,

```

\rRetrieving image: metadata: 13% (93.34kB/s)\r \rRetrieving image: 
metadata: 14% (96.15kB/s)\r \rRetrieving image: metadata: 14% 
(96.15kB/s)\r \r' * Command standard error output: b'Error: Failed 
instance creation: read tcp 192.168.0.5:49524->185.125.190.40:443: read: 
connection reset by peer\n'

```

I'll have to leave that for tomorrow or give you the honor. :P

sergio-costas:
I'll do it.

3v1n0:
> It should be a requirement :-D I use lxcraft to build it, with these changes in the snapcraft.yaml:

Can we make a CI job that handles this a bit like autopkgtests?

So whenever we push to main (at least) or a PR is approved, we check if some snaps are building fine.

sergio-costas:
I'm preparing a patch for the runtime that allows to easily build it from a local SDK snap.

3v1n0:
Good, then can we also use it in CI?

sergio-costas:
Yes, that's my idea.

3v1n0:
Tests passes, so let's go with this in!

Ideally we should also try to build apps with the SDK in CI, but this is fine for now I think.

sergio-costas:
@3v1n0 That's a great idea, too.

3v1n0:
> @3v1n0 That's a great idea, too.

Yeah, I won't ask you to go further this road unless you want to though (in case with a super simple app only), unless you're enjoying with fighting with CI even more

PR #300, opened 2025-05-02
Remove invalid pkgconfig files and broken links

sergio-costas:
Some soft links to libraries that are already in Core 24, and thus, don't needed, are kept in the SDK.

Also, two pkgconfig files are also kept, pointing to the SDK library folder. But since those two libraries (libcrypt and libpciaccess) are already in Core24, and the libraries themselves are deleted, this breaks any build that tries to use any of those.

Removing the broken soft links and the pkgconfig files fixes this. With this patch, installing the development .deb packages using the "build-packages" statement in the snapcraft.yaml ensures that the requires build files are available and that no other pkgconfig file will override the correct ones.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1666,6 +1666,16 @@ parts:
       rm -f usr/lib/*.la
       rm -f usr/lib/*/*.la
 
+      # cleanup broken links; all these are from libc6-dev, libcrypt-dev and libpciaccess-dev
+      for FILE in libcrypt pciaccess libBrokenLocale libanl libc_malloc_debug libmvec libnss_compat libnss_hesiod libpciaccess libresolv libthread_db; do
+        rm -f usr/lib/${CRAFT_ARCH_TRIPLET_BUILD_FOR}/${FILE}.so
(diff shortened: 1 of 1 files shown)


3v1n0:
Thanks for looking into it, but I'd rather prefer continue with #294, I'll double-check if we should drop the files or duplicate them again soon.

sergio-costas:
As I already commented, I think that the key point is in the pkgconfig files that shouldn't be there.

snapcraft.yaml
```diff
@@ -1666,6 +1666,16 @@ parts:
       rm -f usr/lib/*.la
       rm -f usr/lib/*/*.la
 
+      # cleanup broken links; all these are from libc6-dev, libcrypt-dev and libpciaccess-dev
```
nteodosio: Although they should indeed be removed, the AR files are not broken links, so could you just make this more accurate in this comment?

nteodosio (approved):

snapcraft.yaml

3v1n0:
Let's close this for now, as #294 is preferred for now.

PR #301, opened 2025-05-02
46: Update libcanberra

jbicha:
It looks Lennart moved his website. This fixes the build.


jbicha:
Never mind. The old URL is working now again

PR #308, opened 2025-05-26
debs, conditioning: Some cleanups

3v1n0:
See the commits for details

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1672,8 +1672,7 @@ parts:
       rm -rf usr/share/doc
       rm -rf usr/share/gtk-doc
       rm -rf usr/share/man
-      rm -rf usr/libexec/*/installed-tests
-      rm -rf usr/libexec/installed-tests
+      find usr -type d -name installed-tests -exec rm -rfv '{}' \; || true
       rm -f usr/lib/*.la
       rm -f usr/lib/*/*.la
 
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -1661,10 +1666,8 @@ parts:
       rm -rf usr/share/doc
       rm -rf usr/share/gtk-doc
       rm -rf usr/share/man
-      rm -rf usr/libexec/*/installed-tests
-      rm -rf usr/libexec/installed-tests
-      rm -f usr/lib/*.la
-      rm -f usr/lib/*/*.la
+      find usr -type d -name installed-tests -exec rm -rfv '{}' \; || true
+      find usr -type f,l -name '*.la' -print -delete;
```
sergio-costas, line 1670: I think that this should be limited to usr/lib...

sergio-costas (changes requested):
Just a little detail.

snapcraft.yaml
sergio-costas: (recursively, I mean)

3v1n0: Oh, I did set to `usr`...

3v1n0: I don't think anyone really needs to have libtool files anywhere TBH, if you want we can limit to `{usr/lib,lib}`, but IMHO there's no point to ship them anywhere, since they would be broken anyways (content should be redacted with updated prefix)

sergio-costas: ok

sergio-costas (approved):
LGTM

PR #309, opened 2025-05-26
glib: Bump to 2.84.2

3v1n0:
It's the newer stable version, and 2.84.1 fixes various relevant issues


sergio-costas (approved):
Looks sensible.

PR #315, opened 2025-06-09
Use upstream libffi

sergio-costas:
Currently, libffi is being taken from a freedesktop repository, which allows to use Meson. Unfortunately, that repo isn't up to date.

This patch uses the true upstream.

Tested on:
- cheese
- chromium
- evince
- eog
- gnome-text-editor
- darktable
- zoom-client
- gnome-system-monitor
- gimp
- kicad
- gnome-clocks
- gnome-calendar
- gnome-recipes
- gnome-mahjongg

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -104,6 +104,7 @@ parts:
     plugin: autotools
     autotools-configure-parameters:
       - --prefix=/usr
+      - --libdir=/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR
     build-environment: *buildenv
     override-pull: |
       craftctl default
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -97,22 +97,19 @@ parts:
       sed -i 's#aclocaldir="#aclocaldir="$CRAFT_STAGE#' $LIBTOOLIZE
 
   libffi:
-    after: [ libtool, meson-deps ]
-    source: https://gitlab.freedesktop.org/gstreamer/meson-ports/libffi.git
-    source-tag: 'meson-3.2.9999.4'
-# ext:updatesnap
-#   version-format:
-#     format: 'meson-%M.%m.%R'
+    after: [ libtool ]
+    source: https://github.com/libffi/libffi.git
+    source-tag: 'v3.5.0'
     source-depth: 1
-    plugin: meson
-    meson-parameters:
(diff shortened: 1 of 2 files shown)


soumyaDghosh:
Hey are you sure the patch is necessary? I did build it from source without any changes required.

soumyaDghosh:
Uhh, I see yesterday there was a new release created that bumped the minimum autoconf version. I have a small request.

snapcraft.yaml
```diff
@@ -97,22 +97,19 @@ parts:
...
-#     format: 'meson-%M.%m.%R'
+    after: [ libtool ]
+    source: https://github.com/libffi/libffi.git
+    source-tag: 'v3.5.0'
     source-depth: 1
-    plugin: meson
-    meson-parameters:
+    plugin: autotools
+    autotools-configure-parameters:
       - --prefix=/usr
```
soumyaDghosh, line 106:
```suggestion
      - --prefix=/usr
      - --libdir=/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR
```

soumyaDghosh (changes requested):
Adding `libdir` to the configuration would be great IMO

sergio-costas:
> Hey are you sure the patch is necessary? I did build it from source without any changes required.

Yes, it is. Version 3.5.0 requires a too new autoconf version. In fact, the patch is just a backport of one of the commits in master.

snapcraft.yaml
```diff
@@ -97,22 +97,20 @@ parts:
...
-      - -Doptimization=3
-      - -Ddebug=true
```
nteodosio, line 111: Why is debug and optimization dropped?

sergio-costas: Ops...

sergio-costas: Mmm... optimization and debug are enabled only in meson-built parts, but not in the autotools ones... In any.

nteodosio (approved):

nteodosio:
The patch makes sense as 24.04 does not have autoconf 2.72, so this is good to me.

sergio-costas:
@soumyaDghosh So... do you approve?

soumyaDghosh:
> @soumyaDghosh So... do you approve?

Hey yes sure!

PR #327, opened 2025-09-02
Allow local build with local sdk

sergio-costas:
The RUNTIME snap is derived from the corresponding SDK snap, but
currently, the SDK is directly downloaded from the store, from
the CANDIDATE branch. This prevents building a RUNTIME snap with
a locally built SDK snap. But this is a very desirable thing to
do because it's the only way of testing a change in the SDK and
ensure that it doesn't break any snap that depends on them.

This patch includes a python script that allows to build the RUNTIME
using a local SDK snap.

This allows to build the RUNTIME locally with a new, still not
uploaded, SDK, and test it.


snapcraft.yaml
```diff
@@ -35,14 +35,27 @@ slots:
...
+        echo Using GNOME-46-2404-SDK candidate from the store
+        snap install --candidate gnome-46-2404-sdk
+        sdk_version=$(cat ${CRAFT_PART_INSTALL}/snap.gnome-46-2404-sdk/manifest.yaml \
         | yq -r '.version')
+      fi
+
+      cp -a /snap/gnome-46-2404-sdk/current/usr $CRAFT_PART_INSTALL/
+      cp -a /snap/gnome-46-2404-sdk/current/etc $CRAFT_PART_INSTALL/
+      cp -a /snap/gnome-46-2404-sdk/current/var $CRAFT_PART_INSTALL/
+      cp -a /snap/gnome-46-2404-sdk/current/lib $CRAFT_PART_INSTALL/
```
nteodosio: Since it was a stage-snap before, would `$CRAFT_STAGE` be more accurate here?

sergio-costas: Not really, because we are in the install override. And also, I think that I already tested overriding stage, and didn't work.

nteodosio (approved):

snapcraft.yaml
```diff
@@ -35,14 +35,27 @@ slots:
 parts:
   gnome-sdk:
     plugin: nil
-    stage-snaps: [ gnome-46-2404-sdk/latest/candidate ]
+    #stage-snaps: [ gnome-46-2404-sdk/latest/candidate ]
```
3v1n0: I feel this is a bit too risky, wondering...

Can't we keep this and then override pull.

Then, if a local file is around, then we do what you do now in build, otherwise we just leave it to snapcraft via `crafctl default`

snapcraft.yaml
```diff
@@ -35,14 +35,27 @@ slots:
...
     build-packages:
       - yq
     override-build: |
       set -eu
       craftctl default
-      sdk_version=$(cat ${CRAFT_PART_INSTALL}/snap.gnome-46-2404-sdk/manifest.yaml \
+      if [ -e $CRAFT_PROJECT_DIR/gnome-46-2404-sdk.snap ]; then
```
3v1n0: Wondering how we could make this a bit more safe... Maybe picking from another directory is better (as you need to have out it there)... But also, given I don't think we can pass env variables or arguments to snapcraft, can we maybe rely on being on debug (verbose) mode?

I mean, I would like to be sure that this happens on a very clear explicit request only

snapcraft.yaml
sergio-costas: There is at least one problem: we will be mixing the new SDK with the old one, so any non-common file in the old one will remain, thus falsifying the results. Only by fully deleting whatever is installed from that snap would be safe. The problem is... that I'm not sure where exactly is it being downloaded: in theory, the name suggests that it is copied directly into the `stage` folder, but on the other hand, the `override-build` script accesses data from the snap that is placed at `$CRAFT_PART_INSTALL`...

Unless there is a way of doing that `stage-snaps` conditional...

snapcraft.yaml
sergio-costas: Is there a way of detecting if we are in verbose mode? Also... are you sure that the mode isn't triggered when building in launchpad/github CI, to have all the info in case something breaks?

But I agree in making it more safe. I'll investigate it more.

sergio-costas:
@3v1n0 Ok, I moved everything into an external script. This should be much safer.

local-build.py
```diff
@@ -0,0 +1,66 @@
...
+    for line in config:
+        if not completed:
+            if line == 'parts:':
+                found_parts = True
+                config_file.write(line)
+                continue
+            if line == '  gnome-sdk:':
+                found_gnome_part = True
+                config_file.write(line)
+                continue
```
nteodosio: This whole block is doing nothing; `found_parts` and `found_gnome_part` are unused.

local-build.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python3
+
+# This script builds the gnome runtime snap using a local SDK snap.
+
+import sys
+import os
+
+base_config = 'snap/snapcraft.yaml'
+modified_config = 'snapcraft.yaml'
+sdk_file = 'gnome-46-2404-sdk.snap'
```
nteodosio: Can't we use argv[1] instead of a hard-coded name?

local-build.py
```diff
@@ -0,0 +1,66 @@
...
+      cp -a /snap/gnome-46-2404-sdk/current/lib $CRAFT_PART_INSTALL/
+""")
+                continue
+            if line.strip() == '#end':
+                jump_over = False
+                completed = True
+                continue
+        if not jump_over:
+            config_file.write(line)
+
```
nteodosio: I think it would be cleaner to remove all of that and just add a
```
build-environment:
  - LOCAL_BUILD: 0
```
to the part, and then toggle the value here. Then move all this code to snapcraft.yaml instead of injecting it there, e.g.:

```
--> snapcraft.yaml <--
    override-build: |
      set -eu
      craftctl default
      if test -n "$LOCAL_BUILD"; then
      snap install --dangerous $CRAFT_PROJECT_DIR/{sdk_file}
      sdk_version=local
      cp -a /snap/gnome-46-2404-sdk/current/usr $CRAFT_PART_INSTALL/
      cp -a /snap/gnome-46-2404-sdk/current/etc $CRAFT_PART_INSTALL/
      cp -a /snap/gnome-46-2404-sdk/current/var $CRAFT_PART_INSTALL/
      cp -a /snap/gnome-46-2404-sdk/current/lib $CRAFT_PART_INSTALL/
      else
      sdk_version=$(cat ${CRAFT_PART_INSTALL}/snap.gnome-46-2404-sdk/manifest.yaml \
        | yq -r '.version')
      fi
```

I reckon it's simpler to understand, avoids the cryptic #begin and #end directives, and avoids breakage when someone eventually decides to edit that part of snapcraft.yaml without having local-build.py in mind.

nteodosio (changes requested):
What is the motivation for `snapcraft.yaml -> snap/snapcraft.yaml`?

local-build.py
sergio-costas: True. Fixed.

snap/snapcraft.yaml
```diff
@@ -41,8 +41,10 @@ parts:
     override-build: |
       set -eu
       craftctl default
+      #begin
```
3v1n0: I would use something named in a clearer way.

local-build.py
3v1n0: Use `CAPITAL_NAMES` for global variables

local-build.py
```diff
@@ -0,0 +1,66 @@
...
+modified_config = 'snapcraft.yaml'
+sdk_file = 'gnome-46-2404-sdk.snap'
+
+# If there is a modified snapcraft.yaml, delete it before re-generating it
+if os.path.exists(f'./{modified_config}'):
+    os.remove(f'./{modified_config}')
+
+if not os.path.exists(f'./{sdk_file}'):
+    print(f'There is no valid "{sdk_file}" file. Aborting.')
+    sys.exit(-1)
```
3v1n0: -1 is likely unvalid (need to be max 256, just use `1`)

local-build.py
sergio-costas: The point is that we must ensure that there is no risk of an external envvar triggering it accidentally... I'll try another way.

local-build.py
```diff
@@ -0,0 +1,66 @@
...
+base_config = 'snap/snapcraft.yaml'
+modified_config = 'snapcraft.yaml'
+sdk_file = 'gnome-46-2404-sdk.snap'
+
+# If there is a modified snapcraft.yaml, delete it before re-generating it
+if os.path.exists(f'./{modified_config}'):
+    os.remove(f'./{modified_config}')
+
+if not os.path.exists(f'./{sdk_file}'):
+    print(f'There is no valid "{sdk_file}" file. Aborting.')
```
3v1n0:
```suggestion
    print(f'There is no valid "{sdk_file}" file. Aborting.', file=sys.stderr)
```

Or similar, just writing by memory now :)

local-build.py
```diff
@@ -0,0 +1,66 @@
...
+                config_file.write(f"""
+      echo Using local $CRAFT_PROJECT_DIR/{sdk_file}
+      snap install --dangerous $CRAFT_PROJECT_DIR/{sdk_file}
+      sdk_version=local
+
+      cp -a /snap/gnome-46-2404-sdk/current/usr $CRAFT_PART_INSTALL/
+      cp -a /snap/gnome-46-2404-sdk/current/etc $CRAFT_PART_INSTALL/
+      cp -a /snap/gnome-46-2404-sdk/current/var $CRAFT_PART_INSTALL/
+      cp -a /snap/gnome-46-2404-sdk/current/lib $CRAFT_PART_INSTALL/
+""")
```
3v1n0: Can't we use yaml parser / writer? It seems too fragile to me.

local-build.py
```diff
@@ -0,0 +1,66 @@
...
+            if line.strip() == '#end':
+                jump_over = False
+                completed = True
+                continue
+        if not jump_over:
+            config_file.write(line)
+
+os.system('snapcraft clean')
+os.system('snapcraft pack -v')
+os.remove(f'./{modified_config}')
```
3v1n0: You can likely use something like bash TRAP here to cleanup on any error, no?

3v1n0 (commented):
Can we add a CI job testing a simple build with the sdk snap that has been built in the same CI pipeline?

snap/snapcraft.yaml
sergio-costas: Nah, removed that. Give me a second...

nteodosio:
What would be an external envvar?

Am 15/10/2025 um 05:17 schrieb ***@***.***:
> @sergio-costas commented on this pull request.
> 
> 
> 
> > +            if line.strip() == '#begin':  
> +                jump_over = True
> +                config_file.write(f"""
> +      echo Using local $CRAFT_PROJECT_DIR/{sdk_file}
> +      snap install --dangerous $CRAFT_PROJECT_DIR/{sdk_file}
> +      sdk_version=local
> +
> +      cp -a /snap/gnome-46-2404-sdk/current/usr $CRAFT_PART_INSTALL/
> +      cp -a /snap/gnome-46-2404-sdk/current/etc $CRAFT_PART_INSTALL/
> +      cp -a /snap/gnome-46-2404-sdk/current/var $CRAFT_PART_INSTALL/
> +      cp -a /snap/gnome-46-2404-sdk/current/lib $CRAFT_PART_INSTALL/
> +""")
> +                continue
> +            if line.strip() == '#end':
> +                jump_over = False
> +                completed = True
> +                continue
> +        if not jump_over:
> +            config_file.write(line)
> +
> 
> The point is that we must ensure that there is no risk of an external envvar
> triggering it accidentally... I'll try another way.
>

local-build.py
sergio-costas: Done.

local-build.py
sergio-costas: Ok.

local-build.py
sergio-costas: Yes, that's how it's done.

local-build.py
sergio-costas: Modified. Now it's much simpler.

local-build.py
sergio-costas: Mmm... not sure what do you mean...

sergio-costas:
@nteodosio Changes done.

local-build.py
sergio-costas: Done.

local-build.py
sergio-costas: Used a different approach...

sergio-costas:
The motivation for snapcraft.yaml -> snap/snapcraft.yaml is because we have to modify it, and since the one in the project folder has priority over the one at snap folder, we can preserve the original one and just create a modified version that will be used over the original one.

sergio-costas:
An external envvar is an evironment variable not defined inside the script, but in the snapcraft.yaml file or when calling snapcraft.

nteodosio:
Oh sorry response by email put it out of thread... I'll respond in the thread.

snap/snapcraft.yaml
```diff
@@ -41,8 +41,21 @@ parts:
     override-build: |
       set -eu
       craftctl default
-      sdk_version=$(cat ${CRAFT_PART_INSTALL}/snap.gnome-46-2404-sdk/manifest.yaml \
-        | yq -r '.version')
+      # This export is replaced by `local-build.py` when using a local snap
+      export LOCAL_SDK_SNAP=""
+      if [[ -z ${LOCAL_SDK_SNAP} ]]; then
```
nteodosio: `[[` will only work with Bash; For POSIX compliance, prefer `[` or `test`.

snap/snapcraft.yaml
```diff
@@ -41,8 +41,21 @@ parts:
     override-build: |
       set -eu
       craftctl default
-      sdk_version=$(cat ${CRAFT_PART_INSTALL}/snap.gnome-46-2404-sdk/manifest.yaml \
-        | yq -r '.version')
+      # This export is replaced by `local-build.py` when using a local snap
+      export LOCAL_SDK_SNAP=""
```
nteodosio: `export` is not needed here as the variable is not being used in a child process.

nteodosio (changes requested):
Thanks, just small remarks.

snap/snapcraft.yaml
sergio-costas: Fixed.

local-build.py
nteodosio: > An external envvar is an evironment variable not defined inside the script, but in the snapcraft.yaml file or when calling snapcraft.

With `X=1 snapcraft pack`, X will not be defined for any snapcraft.yaml scriplet, so clashing that way is not possible.

Now, in snapcraft.yaml itself, I'd hope someone altering it would check if the variable wasn't used before? In any case LOCAL_BUILD is quite a clear meaning variable.

Though the new way, with LOCAL_SDK_SNAP, is alright to me as well.

nteodosio:
By the way, I'm opening https://launchpad.net/bugs/2127927 so that this can eventually addressed where it should be, namely Snapcraft itself.

snap/snapcraft.yaml
sergio-costas: Done.

nteodosio (approved):

sergio-costas:
@3v1n0 Hey, can you check this when you have some spare time, please? I'm nearly at the point where my autoupdater for snaps will require this to automatically test updates for the SDK...

local-build.py
```diff
@@ -0,0 +1,40 @@
...
+    sys.exit(1)
+
+with open(f'./{BASE_CONFIG}', "r") as config_file:
+    config = config_file.readlines()
+
+with open(f'./{MODIFIED_CONFIG}', "w") as config_file:
+    config_file.write('# DO NOT EDIT THIS FILE. IT IS AUTOMATICALLY GENERATED\n')
+    config_file.write('# Edit snap/snapcraft.yaml and re-run the local-build.py script\n\n')
+    for line in config:
+        if line.strip().startswith('stage-snaps:'):
```
3v1n0: I'd prefer to use python yaml to parse things here

local-build.py
```diff
@@ -0,0 +1,40 @@
...
+BASE_CONFIG = 'snap/snapcraft.yaml'
+MODIFIED_CONFIG = 'snapcraft.yaml'
+SDK_FILE = 'gnome-46-2404-sdk.snap'
+
+if len(sys.argv) > 1:
+    SDK_FILE = sys.argv[1]
+
+# If there is a modified snapcraft.yaml, delete it before re-generating it
+if os.path.exists(f'./{MODIFIED_CONFIG}'):
+    os.remove(f'./{MODIFIED_CONFIG}')
```
3v1n0, line 26: I mean can't we just put the whole code inside a `try...finally` statement and cleanup on `finally`?

local-build.py
```diff
@@ -0,0 +1,40 @@
...
+
+import sys
+import os
+
+BASE_CONFIG = 'snap/snapcraft.yaml'
+MODIFIED_CONFIG = 'snapcraft.yaml'
+SDK_FILE = 'gnome-46-2404-sdk.snap'
+
+if len(sys.argv) > 1:
+    SDK_FILE = sys.argv[1]
```
3v1n0: nit, use inverse logic:


```suggestion
sdk_file = sys.argv[1] if len(sys.argv) > 1 else SDK_FILE
```

And then use `sdk_file` around

snap/snapcraft.yaml
```diff
@@ -41,8 +41,22 @@ parts:
     override-build: |
       set -eu
       craftctl default
-      sdk_version=$(cat ${CRAFT_PART_INSTALL}/snap.gnome-46-2404-sdk/manifest.yaml \
-        | yq -r '.version')
+      # This export is replaced by `local-build.py` when using a local snap
+      LOCAL_SDK_SNAP=""
+      if [ -z ${LOCAL_SDK_SNAP} ]; then
```
3v1n0: Can't we just use `if [ -z "${LOCAL_SDK_SNAP:-}" ]; then`

In any case, quotes are necessary.

snap/snapcraft.yaml
```diff
@@ -41,8 +41,22 @@ parts:
...
+          | yq -r '.version')
+      else
+        echo Using local $CRAFT_PROJECT_DIR/${LOCAL_SDK_SNAP}
+        snap install --dangerous $CRAFT_PROJECT_DIR/${LOCAL_SDK_SNAP}
+
+        cp -a /snap/gnome-46-2404-sdk/current/usr $CRAFT_PART_INSTALL/
+        cp -a /snap/gnome-46-2404-sdk/current/etc $CRAFT_PART_INSTALL/
+        cp -a /snap/gnome-46-2404-sdk/current/var $CRAFT_PART_INSTALL/
+        cp -a /snap/gnome-46-2404-sdk/current/lib $CRAFT_PART_INSTALL/
+        sdk_version=46
```
3v1n0: Can't we parse this value instead of hardcoding it?

local-build.py
3v1n0: I meant https://github.com/ubuntu/gnome-sdk/pull/327#discussion_r2675621477 but let's handle it there.

3v1n0:
> Can we add a CI job testing a simple build with the sdk snap that has been built in the same CI pipeline?

Is this doable

snap/snapcraft.yaml
sergio-costas: No, we can't. The python script modifies the definition to put instead the path. But the quotes... yes, that makes sense.

local-build.py
sergio-costas: In fact, it's not a good idea to do this because only files in the current folder and below are accessible from inside `snapcraft`, and pointing to a .snap placed outside makes the script fail. So I'll remove this.

local-build.py
sergio-costas: This is to ensure that any old snap file is removed before doing anything here.

local-build.py
sergio-costas: Done, although it required some tricks to ensure that the output did work.

snap/snapcraft.yaml
sergio-costas: The problem is that the `manifest.yaml` file seems to be added by the store, it isn't available inside the locally built snaps...

3v1n0: I mean do not we have a way to figure out that we're in 46 or even in `gnome-46-2404-sdk`?

Not a big deal though, but i'd just add a single variable used in both `if` branches.

local-build.py
```diff
@@ -21,19 +30,25 @@
...
+# remove the stage-snaps entry
+del gnome_part['stage-snaps']
+
+build_script = gnome_part['override-build']
+new_script = ""
+for line in build_script.splitlines():
+    if line.startswith('LOCAL_SDK_SNAP='):
+        pos = line.find('=')
+        line = f'{line[:pos]}="{SDK_FILE}"'
+    new_script += line + '\n'
```
3v1n0: One way to avoid this, and just do it in the whole yaml, could be setting it as value in `parts.buildenv.build-environment`, no?

So we define it from there only in this case, without having to add some fragile bash syntax that this script modifies

3v1n0: Oh, we do not have that in the runtime side, well, can't be still as `parts.gnome-sdk.build-environment`

snap/snapcraft.yaml
sergio-costas: I can use the python script to use the `name` entry for that...

local-build.py
sergio-costas: Oh! Of course... that's really elegant!

snap/snapcraft.yaml
sergio-costas: Done, also using the build-environment.

local-build.py
```diff
@@ -37,15 +37,8 @@ def str_presenter(dumper, data):
...
-new_script = ""
-for line in build_script.splitlines():
-    if line.startswith('LOCAL_SDK_SNAP='):
-        pos = line.find('=')
-        line = f'{line[:pos]}="{SDK_FILE}"'
-    new_script += line + '\n'
-
-gnome_part['override-build'] = new_script
+version = config['name'].split('-')[1]
+gnome_part['build-environment'] = [{'LOCAL_SDK_SNAP': SDK_FILE}, {'sdk_version': version}]
```
3v1n0: nit:

```python
gnome_part['build-environment'] = [
    {'LOCAL_SDK_SNAP': SDK_FILE},
    {'sdk_version': version},
]
```

snap/snapcraft.yaml
```diff
@@ -41,8 +41,20 @@ parts:
     override-build: |
       set -eu
       craftctl default
-      sdk_version=$(cat ${CRAFT_PART_INSTALL}/snap.gnome-46-2404-sdk/manifest.yaml \
-        | yq -r '.version')
+      # This export is replaced by `local-build.py` when using a local snap
```
3v1n0:
```suggestion
      # This is set by `local-build.py` when using a local snap
```

sergio-costas: Done.

local-build.py
sergio-costas: Done.

sergio-costas:
> Can we add a CI job testing a simple build with the sdk snap that has been built in the same CI pipeline?

You mean to build gnome-sdk, and then build gnome-runtime with it?

Or, in the gnome-sdk branch, build both?

local-build.py
3v1n0: Ah let's also be future proof in case we'll set `gnome_part['build-environment']`, so use something like:

`nome_part['build-environment'] = gnome_part.get('build-environment', []).extend(...)`

3v1n0:
> You mean to build gnome-sdk, and then build gnome-runtime with it?

Yes.

If we really want also in the runtime branch we could do it, but I'd build the sdk, once it's done we build the runtime with it, and save as artifact so that it can be tested locally.

Ideally we could even do some smoke-test of an app using it in CI, but you don't have to go that far for now I think

sergio-costas:
> Ideally we could even do some smoke-test of an app using it in CI, but you don't have to go that far for now I think

That's what I'm going to do with snap-udpater-go ;-)

local-build.py
sergio-costas: Done, although not with `extend` (that does it in-place)

sergio-costas:
@3v1n0 So... can I merge this, then?

3v1n0:
@sergio-costas I was wondering if we could get that workflow in this branch too, so we can merge both the change and check it works in the same PR

local-build.py
3v1n0: Fine, but please still use the `try...finally` syntax for the rest.

sergio-costas:
Mmm... but if the idea is to build the SDK and the runtime at the same time, then we have to merge this first, and then modify the CI in the SKD branch...

sergio-costas:
(accidentally pressed "close with comment" instead of "comment")

3v1n0:
> Mmm... but if the idea is to build the SDK and the runtime at the same time, then we have to merge this first, and then modify the CI in the SKD branch...

Oh, right, I keep forgetting this is targetting the runtime side only :-D

local-build.py
sergio-costas: Something like this...?

sergio-costas: (at the end of the file)

3v1n0 (approved):

PR #328, opened 2025-09-19
Update to (kind-of) last versions

sergio-costas:
Updated several libraries to more recent/latest versions.

Tested with

* cheese
* darktable
* element-desktop (fails, but seems to be a problem in the element snap)
* eog
* epiphany
* evince
* folio (fails, but it's a problem in the snap because fails with stable too. Folio edge version works)
* gimp
* gnome-characters
* gnome-mahjongg
* gnome-mines
* gnome-recipes
* gnome-system-monitor
* gnome-text-editor
* kicad
* mattermost-desktop
* shotwell
* telegram-desktop

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -65,7 +65,7 @@ parts:
     after: [ ninja ]
     plugin: nil
     source: https://github.com/mesonbuild/meson.git
-    source-tag: '1.8.1'
+    source-tag: '1.8.4'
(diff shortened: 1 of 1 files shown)


kenvandine (approved):
I don't see anything risky there, looks good to me!

sergio-costas:
@kenvandine Except that it's Friday! 😅 Let's wait to Monday.

PR #329, opened 2025-09-26
Pull early, not in build or prime phases

nteodosio:
In Launchpad, Risc builds fail when Meson attempts to pull subprojects late.

Build attempts at https://launchpad.net/~nteodosio/+snap/sdk-46-pull-early-for-risc.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -108,6 +108,7 @@ parts:
     build-environment: *buildenv
     build-packages:
       - texinfo
+      - libltdl-dev
 
   glib:
     after: [ libffi, meson-deps ]
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -839,7 +839,6 @@ parts:
     override-pull: |
       set -eux
       craftctl default
-      meson subprojects download
       sed -i 's#Werror=missing-include-dirs#Wmissing-include-dirs#g' meson.build
 
   libdazzle:
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -860,35 +860,8 @@ parts:
       craftctl default
       sed -i 's#Werror=missing-include-dirs#Wmissing-include-dirs#g' meson.build
 
-  libcanberra:
-    after: [ libdazzle ]
-    source: http://0pointer.de/lennart/projects/libcanberra/libcanberra-0.30.tar.xz
-# ext:updatesnap
-#   version-format:
-#     ignore: true
-    plugin: autotools
-    autotools-configure-parameters:
-      - --prefix=/usr
-      - --with-builtin=pulse
-    build-environment: *buildenv
-    override-pull: |
-      craftctl default
-      for p in $CRAFT_PROJECT_DIR/patches/libcanberra/*.patch; do
-        patch -p1 < "$p"
-      done
-    override-build: |
-      ./autogen.sh
-      craftctl default
-    build-packages:
-      - libasound2-dev
-      - libvorbis-dev
-      - libtdb-dev
-      - libpulse-dev
-      - libgstreamer1.0-dev
-      - libltdl-dev
-
   gsound:
-    after: [ libcanberra, meson-deps, vala, gobject-introspection ]
+    after: [ meson-deps, vala, gobject-introspection ]
     source: https://gitlab.gnome.org/GNOME/gsound.git
     source-tag: '1.0.3'
     source-depth: 1
@@ -901,6 +874,9 @@ parts:
       - -Dintrospection=true
       - -Denable_vala=true
     build-environment: *buildenv
+    build-packages:
+      - libcanberra-dev
+      - libcanberra-gtk3-dev
 
   gsettings-desktop-schemas:
     after: [ gsound, meson-deps, gobject-introspection ]
@@ -1422,6 +1398,7 @@ parts:
       - libbz2-dev
       - libcairo2-dev
       - libcanberra-gtk3-dev
+      - libcanberra-pulse
       - libcrypt-dev
       - libcurl4-openssl-dev
       - libdbus-1-3
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1360,13 +1360,23 @@ parts:
 
   debs:
     after:
+      - atkmm
       - at-spi2-core
+      - cairomm
+      - glibmm
+      - gtk-locales
+      - gtkmm
+      - gtksourceview
       - libadwaita
       - libayatana-appindicator
+      - libdazzle
       - libgnome-games-support
       - libgweather
+      - libportal
       - librest
       - libsoup3
+      - mm-common
+      - pangomm
       - webp-pixbuf-loader
     plugin: nil
     stage-packages:
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -372,9 +372,11 @@ parts:
     build-environment: *buildenv
     build-packages:
       - libssl-dev
+      - meson
       - curl
     override-pull: |
       craftctl default
+      meson subprojects download
       # cargo version in .deb is too old
       curl https://sh.rustup.rs -sSf > cargo.sh
       sh ./cargo.sh -y
@@ -434,9 +436,11 @@ parts:
     build-packages:
       - libidn2-dev
       - libunistring-dev
+      - meson
     override-pull: |
       set -eux
       craftctl default
+      meson subprojects download
       sed -i 's#^\#!/usr/bin/env python$#\#!/usr/bin/env python3#g' src/psl-make-dafsa
 
   libsoup3:
@@ -538,8 +542,10 @@ parts:
       - libxi-dev
       - libxkbfile-dev
       - libxml2-utils
+      - meson
     override-pull: |
       craftctl default
+      meson subprojects download
       patch -p1 < $CRAFT_PROJECT_DIR/patches/gdkglcontext-wayland-Fallback-to-GLES-2_0-after-GL-failed.patch
 
   gtk4:
@@ -632,8 +638,10 @@ parts:
       - libsystemd-dev
       - gperf
       - libappstream-dev
+      - meson
     override-pull: |
       craftctl default
+      meson subprojects download
       patch -p1 < $CRAFT_PROJECT_DIR/patches/stylesheet-common-Use-disabled-opacity-for-disabled-pictu.patch
       patch -p1 < $CRAFT_PROJECT_DIR/patches/stylesheet-common-Use-opacity-for-all-disabled-gtk-images.patch
       patch -p1 < $CRAFT_PROJECT_DIR/patches/accent-color-Return-return-orange-as-default-color.patch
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -851,35 +862,8 @@ parts:
       craftctl default
       sed -i 's#Werror=missing-include-dirs#Wmissing-include-dirs#g' meson.build
 
-  libcanberra:
-    after: [ libdazzle ]
-    source: http://0pointer.de/lennart/projects/libcanberra/libcanberra-0.30.tar.xz
```
3v1n0, line 856: You could also just use salsa or https://git.launchpad.net/ubuntu/+source/libcanberra/log/?h=applied/ubuntu/noble-devel that has all the patches set too.

3v1n0 (approved):
See if can still use complied libcanberra or if not needed.

Oterwise, lgtm

snapcraft.yaml
nteodosio: > See if can still use complied libcanberra or if not needed.

I don't understand, can you please rephrase? Canberra is needed for the Gsound part.

nteodosio:
Build succeeded in Launchpad: https://launchpad.net/~nteodosio/+snap/sdk-46-pull-early-for-risc/+build/2926474

snapcraft.yaml
3v1n0: Instead of using 0pointer, use the launchpad or salsa repositories with their patches

nteodosio: So just ship the debs?

3v1n0: Kinda, but not sure if @sergio-costas had another rationale for building it from source.

kenvandine (commented):
Do we need to add meson to build-packages? We have a meson-deps part that should give us meson.

nteodosio:
> Do we need to add meson to build-packages? We have a meson-deps part that should give us meson.

That must indeed be redundant; Removed.

Though do we really need to build meson from source instead of using the one from Ubuntu's archive?

The inclusion of Meson as a part I traced back to

```
commit 29ddf56a29c414e3e0418f112ff181bfe9185e90
Author: Olivier Tilloy <olivier.tilloy@canonical.com>
Date:   Thu Mar 15 17:25:57 2018 +0100

    Various fixes to make all parts build on 16.04:

     - add a meson part (json-glib requires 0.44)
     - add a missing build dep for vala (libgraphviz-dev)
     - do not alter $VAPIDIR and $VALAFLAGS when building the vala part
```

That initial motivation is probably no longer valid.

kenvandine:
> > Do we need to add meson to build-packages? We have a meson-deps part that should give us meson.
> 
> That must indeed be redundant; Removed.
> 
> Though do we really need to build meson from source instead of using the one from Ubuntu's archive?
> 
> The inclusion of Meson as a part I traced back to
> 
> ```
> commit 29ddf56a29c414e3e0418f112ff181bfe9185e90
> Author: Olivier Tilloy <olivier.tilloy@canonical.com>
> Date:   Thu Mar 15 17:25:57 2018 +0100
> 
>     Various fixes to make all parts build on 16.04:
> 
>      - add a meson part (json-glib requires 0.44)
>      - add a missing build dep for vala (libgraphviz-dev)
>      - do not alter $VAPIDIR and $VALAFLAGS when building the vala part
> ```
> 
> That initial motivation is probably no longer valid.

That's a good question. At some point, snapcraft started requiring a meson-deps part whenever using the meson plugin, but that might have changed.

nteodosio:
@kenvandine, after that commit the build in the CI failed with:

```
Detailed information: 
:: + craftctl default
:: + meson subprojects download
:: /tmp/tmpwujaa0fs/scriptlet.sh: line 6: meson: command not found
Recommended resolution: Review the scriptlet and make sure it's correct.
For more information, check out: https://documentation.ubuntu.com/snapcraft/8.12.0/reference/plugins/
Failed to run snapcraft in instance
Full execution log: '/home/runner/.local/state/snapcraft/log/snapcraft-20251013-132514.336669.log'
Error: The process '/usr/bin/sudo' failed with exit code 1
```

So I'm reverting it.

nteodosio:
(For the Librsvg part.)

kenvandine:
> @kenvandine, after that commit the build in the CI failed with:
> 
> ```
> Detailed information: 
> :: + craftctl default
> :: + meson subprojects download
> :: /tmp/tmpwujaa0fs/scriptlet.sh: line 6: meson: command not found
> Recommended resolution: Review the scriptlet and make sure it's correct.
> For more information, check out: https://documentation.ubuntu.com/snapcraft/8.12.0/reference/plugins/
> Failed to run snapcraft in instance
> Full execution log: '/home/runner/.local/state/snapcraft/log/snapcraft-20251013-132514.336669.log'
> Error: The process '/usr/bin/sudo' failed with exit code 1
> ```
> 
> So I'm reverting it.

Perhaps snapcraft uses meson directly rather than by executing the meson command. The meson command is probably somewhere in $CRAFT_STAGE.

nteodosio:
As can be seen from the last two commits, even using
`$CRAFT_STAGE/usr/bin/meson` fails:

> :: /tmp/tmpoea6p4h8/scriptlet.sh: line 6: /root/stage/usr/bin/meson: No such
file or directory

And I now understand why: The pull phases are run even before parts specified in
`after` have run. So meson-deps did not yet run when Librsvg pull runs, and
thus it finds no meson.

Confusingly, even using the corresponding plug-in does not give access to the
command. This is something we encountered in Firefox, in which a git plug-in
would not give access to the `git` command in an override-* script; not without
adding git to build-packages.

So I think the simplest solution is indeed to return to the original state,
namely with meson speficied as a build-package for each introduced `meson
download`.

Am 13/10/2025 um 14:11 schrieb ***@***.***:
> kenvandine left a comment (ubuntu/gnome-sdk#329)
> 
> > @kenvandine, after that commit the build in the CI failed with:
> > 
> > ```
> > Detailed information: 
> > :: + craftctl default
> > :: + meson subprojects download
> > :: /tmp/tmpwujaa0fs/scriptlet.sh: line 6: meson: command not found
> > Recommended resolution: Review the scriptlet and make sure it's correct.
> > For more information, check out:
> > https://documentation.ubuntu.com/snapcraft/8.12.0/reference/plugins/ Failed
> > to run snapcraft in instance Full execution log:
> > '/home/runner/.local/state/snapcraft/log/snapcraft-20251013-132514.336669.log'
> > Error: The process '/usr/bin/sudo' failed with exit code 1 ```
> > 
> > So I'm reverting it.  
> 
> Perhaps snapcraft uses meson directly rather than by executing the meson
> command. The meson command is probably somewhere in $CRAFT_STAGE.
>

snapcraft.yaml
sergio-costas: It was already that way when I arrived... 😇

snapcraft.yaml
```diff
@@ -1665,6 +1659,9 @@ parts:
       - python3-pip
       - zip
       - python3-apt
+    override-pull: |
+      craftctl default
+      python3 -m pip install --break-system-packages pyyaml pyelftools
```
sergio-costas, line 1667: What about using the deb packages?

snapcraft.yaml
```diff
@@ -434,9 +439,11 @@ parts:
     build-packages:
       - libidn2-dev
       - libunistring-dev
+      - meson
```
sergio-costas: Build packages are added during pull, and are installed in the system, so it's not needed to duplicate meson.

snapcraft.yaml
3v1n0: I see the only rationale it may be to make sure that it's able to find resources in other paths other than /usr, but I'm unsure if anything is needed.

However, it's a really old module that not sure anyone cares about nowadays.

So let's just rely on deb for now.

snapcraft.yaml
sergio-costas: I think that in core22 they weren't available, and that's why I used it, but in core24 they are...

sergio-costas (changes requested):
Some nitpicks

sergio-costas:
@nteodosio But, AFAIK, the build-packages are installed before `pull` is done... Have you tried to install meson as a build-packages in the first part?

snapcraft.yaml
nteodosio: As per

- https://launchpad.net/ubuntu/+source/python-pyelftools
- https://launchpad.net/ubuntu/+source/pyyaml

They were available since 16.04.

I'm all for debs, I'd be happy to follow up with a merge request for that though, but would prefer to leave this as is now so as to not diverge too much.

snapcraft.yaml

nteodosio: Ah, that's news for me. But what if one of a part is removed, then suddenly another part using Meson will fail. Is it not better to be redundant here?

nteodosio:
Is "first part" well defined? I'm afraid relying on implicit ordering is bound to cause unexpected errors down the line.

sergio-costas:
That's the reason for the "after:" statement in the snapcraft.yaml file. If you ensure that any part that uses `meson` depends on the part that installs it, it works.

sergio-costas:
So... maybe replacing "meson-deps" build with a `build-packages` that installs meson... 🤔

nteodosio:
Ah, so with first part you mean not the first part to run in the build, but `meson-deps`?

sergio-costas:
Well... meson-deps isn't really the first part, because `ninja` is built before, but... yes, the idea is to ensure that before any other part, the one that installs meson must be run.

sergio-costas:
And the dependencies are broken: you must ensure that the "DEBS" part has an "after" entry that guarantees that it is built after those parts that are listed in the `test-dependencies` test.

snapcraft.yaml
sergio-costas: Good point.

snapcraft.yaml
```diff
@@ -373,8 +375,11 @@ parts:
     build-packages:
       - libssl-dev
       - curl
+      - cargo-1.82
     override-pull: |
       craftctl default
+      meson subprojects download
+      cargo-1.82 update
```
sergio-costas: Mmm... but, if we are already downloading cargo just after this, why update it? I mean: if installing cargo-1.82 and calling `update` is enough, we can remove the next lines, isn't it?

snapcraft.yaml
```diff
@@ -851,35 +858,8 @@ parts:
       craftctl default
       sed -i 's#Werror=missing-include-dirs#Wmissing-include-dirs#g' meson.build
 
-  libcanberra:
```
sergio-costas, line 854: Maybe we should also add the libcanberra-pulse package, since this was built with pulseaudio... what do you think?

sergio-costas (commented):
Two little nitpicks, but otherwise it looks great.

snapcraft.yaml
nteodosio: Add where? Isn't it only a Gsound build dependency?

sergio-costas: No, it is already being added in `debs` part (`libcanberra-gtk3-dev`, which adds everything else)

snapcraft.yaml
sergio-costas: Ok, it seems that it doesn't work that way...

nteodosio: Now it does.

sergio-costas: Love it!

sergio-costas (approved):
Now it looks perfect.

PR #330, opened 2025-10-01
Fix python version

sergio-costas:
The python version defined in snapcraft-desktop-integration is 3.10, which is wrong. The right value is 3.12.

Since the value should be overrode using Make parameters, this patch fixes this by specifying the right value when building the command chain.


kenvandine (approved):
Looks good, thanks!

PR #331, opened 2025-10-17
Fix late crate download for librsvg part on RISC-V.

nteodosio:


snapcraft.yaml
```diff
@@ -374,15 +374,22 @@ parts:
...
+      # launchpad.net/~desktop-snappers/+snap/gnome-46-2404-sdk/+build/2931431
+      # shows it does not.
+      cargo-1.82 update
       # cargo version in .deb is too old
       curl https://sh.rustup.rs -sSf > cargo.sh
       sh ./cargo.sh -y
       export PATH=$PATH:$HOME/.cargo/bin
-      ~/.cargo/bin/cargo update
       ~/.cargo/bin/cargo install cargo-c --locked
       cp ~/.cargo/bin/* /usr/local/bin
```
sergio-costas, line 396: What's the point on downloading cargo as deb, and then download the last version too, but not update that last one?

nteodosio: What do you mean with 'update that last one'?

Librsvg does depend on way too new Cargo (1.85).

sergio-costas: Ok, after what you commented in the meeting, it's a trick...

sergio-costas (approved):
I dislike this hack, but it seems that it's the only way to have RiscV building... But what do you say about enabling it only for RiscV, checking environment variables?

sergio-costas:
Mmm... wait, I have some ideas... give me some minutes to test them...

sergio-costas:
Nah... didn't work. I don't know why it doesn't recognize the already compiled crates :-(

snapcraft.yaml
```diff
@@ -374,15 +374,24 @@ parts:
...
+      # launchpad.net/~desktop-snappers/+snap/gnome-46-2404-sdk/+build/2931431
+      # shows it does not.
+      if test "$CRAFT_ARCH_BUILD_FOR" = riscv64; then
+        cargo-1.82 update
+      fi
       # cargo version in .deb is too old
       curl https://sh.rustup.rs -sSf > cargo.sh
       sh ./cargo.sh -y
       export PATH=$PATH:$HOME/.cargo/bin
-      ~/.cargo/bin/cargo update
```
sergio-costas, line 385: I would add a "if not test..." to run this on non-riscv arches.

nteodosio: This runs on non-riscv arches.

sergio-costas: Is it being already run when calling "cargo.sh"?

nteodosio: I do not know; Why is it relevant?

sergio-costas: I mean... we had it before, if we remove that line...

nteodosio: Oh sorry you mean you want to keep running
```
       ~/.cargo/bin/cargo update
```
on non-RISCV, of course, let me add it.

nteodosio: Though you know that line was added by #329, right? So in essence it was not needed in those architectures.

sergio-costas: Oh, good point! Then leave it as is.

PR #334, opened 2026-01-08
Add fontconfig with nosleep patch

sergio-costas:
Added fontconfig with the nosleep patch, that includes a new option that disables the two-seconds delay at the end of the fc-cache command.

This is to reduce the seeding time in Ubuntu.

https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/2116949/

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -107,8 +107,6 @@ parts:
       - --prefix=/usr
       - --libdir=/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR
     build-environment: *buildenv
-    build-packages:
-      - libfontconfig1-dev
     override-pull: |
       craftctl default
       patch -p1 < $CRAFT_PROJECT_DIR/patches/fontconfig-delay.patch
@@ -1589,6 +1587,8 @@ parts:
       rm -f $CRAFT_PART_INSTALL/usr/lib/*/libatk-1.0.so*
       rm -f $CRAFT_PART_INSTALL/usr/lib/*/libatk-bridge-2.0.so*
       rm -f $CRAFT_PART_INSTALL/usr/lib/*/libatspi.so*
+    stage:
+      - -usr/bin/fc-cache
 
   conditioning:
     after:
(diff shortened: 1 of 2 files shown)


snapcraft.yaml
```diff
@@ -1387,6 +1405,7 @@ parts:
       - mm-common
       - pangomm
       - webp-pixbuf-loader
+      - fontconfig
```
nteodosio, line 1406: Don't we want
```
- -usr/bin/fc-cache
```
here too like the other merge request?

sergio-costas: It seems to not be needed... but I'll add just in case...

sergio-costas: Done.

nteodosio (approved):

sergio-costas:
@nteodosio This is failing due to not enough disk space... Do you know if there is a way of increasing the size? If not, I think that I have an idea to slightly reduce the disk usage, but it's a hack...

3v1n0:
> @nteodosio This is failing due to not enough disk space... Do you know if there is a way of increasing the size? If not, I think that I have an idea to slightly reduce the disk usage, but it's a hack...

Is it due to the container size or the actual github container?

If it's a container thing, then maybe there's some snapcraft option (I didn't dig too much), otherwise an option would be to switch to the canonical builders

nteodosio:
Sorry I missed your question. Yeah the question is: Where is it failing? Launchpad, Github, local? I don't know a snapcraft option to do that, but you can increase pool storage size in Lxd.

All else failing, there are some hacks, but then I need to see have the log to see where it's failing.

sergio-costas:
> All else failing, there are some hacks, but then I need to see have the log to see where it's failing.

It's failing here, when doing the CI in Github.

sergio-costas:
@nteodosio These are the logs https://github.com/ubuntu/gnome-sdk/actions/runs/20848439086/job/59897493951 and https://github.com/ubuntu/gnome-sdk/actions/runs/20848237007/job/59896833471

sergio-costas:
Anyway, I'm building it locally and it works, so I think that it confirms that the problem is in github...

nteodosio:
If it passes on Launchpad, then we're good, otherwise we have to look into
removing $PART_CRAFT_SRC for parts already built.

PR #335, opened 2026-01-08
Add fontconfig with nosleep patch 2204

sergio-costas:
Added fontconfig with the nosleep patch, that includes a new option that disables the two-seconds delay at the end of the fc-cache command.

This is to reduce the seeding time in Ubuntu.

https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/2116949/

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -100,8 +100,6 @@ parts:
       - --prefix=/usr
       - --libdir=/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR
     build-environment: *buildenv
-    build-packages:
-      - libfontconfig1-dev
     override-pull: |
       craftctl default
       patch -p1 < $CRAFT_PROJECT_DIR/patches/fontconfig-delay.patch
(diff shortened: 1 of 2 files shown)


snapcraft.yaml
```diff
@@ -90,6 +90,23 @@ parts:
...
+    source: https://gitlab.freedesktop.org/fontconfig/fontconfig.git
+    source-tag: 2.13.1
+    source-depth: 1
+    plugin: autotools
+    autotools-configure-parameters:
+      - --prefix=/usr
+      - --libdir=/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR
+    build-environment: *buildenv
+    build-packages:
+      - libfontconfig1-dev
```
nteodosio: Is this really necessary?

nteodosio (approved):

snapcraft.yaml
sergio-costas: Good point... anyway, it is already being added in other places,so...

sergio-costas: Removed.

3v1n0 (approved):

PR #332, opened 2025-12-18
feat: Use mesa 25.2 backported to core22

kenvandine:
This includes Mesa 25.2.8 from a backport PPA which enables more recent Intel hardware and includes significant battery usage improvements for more recent AMD hardware.

https://launchpad.net/~desktop-snappers/+archive/ubuntu/core22-mesa-backports

PR #337, opened 2026-01-08
fix: Make sure we pull from the same backport PPA as the sdk snap

kenvandine:
Since the gnome-42-2204-sdk snap is using the Mesa backport for core22 in a PPA, we should ensure the runtime matches.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -32,6 +32,10 @@ slots:
       read:
         - /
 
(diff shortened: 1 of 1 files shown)


nteodosio (approved):
Given #332 this looks correct to me.

sergio-costas (approved):
Looks sensible.

PR #339, opened 2026-01-09
snapcraft: Include latest stable libnotify version

3v1n0:
Libnotify has code that is snap-related upstream so let's use the latest version to ensure we are in sync with latest snapd expectations

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1290,6 +1290,24 @@ parts:
     build-packages:
       - gir1.2-glib-2.0
 
+  libnotify:
+    after: [ glib, gobject-introspection, gdk-pixbuf ]
+    source: https://gitlab.gnome.org/GNOME/libnotify.git
+    source-tag: '0.8.8'
+    source-depth: 1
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -1290,6 +1290,23 @@ parts:
...
+    source-depth: 1
+    plugin: meson
+    build-environment: *buildenv
+    meson-parameters:
+      - --prefix=/usr
+      - -Doptimization=3
+      - -Dintrospection=enabled
+      - -Dgtk_doc=false
+      - -Dtests=false
+      - -Dman=false
```
sergio-costas, line 1307: I would add this

```suggestion
      - -Dman=false
      - -Ddocbook_docs=disabled
```

sergio-costas (approved):
LGTM. Just a simple suggestion, but completely optional.

snapcraft.yaml
3v1n0: Done!

PR #340, opened 2026-01-09
snapcraft: Remove sources and built binaries at stage phase

sergio-costas:
A test deleting sources. DO NOT MERGE!!!!


3v1n0:
Not great, but if it can unblock for us it's fine, maybe just add in all these lines a `FIXME:` comment mentioning that this should be dropped once we're done.

Technically saving space is not a big thing in general, but this could definitely lead to undefined beahvior

PR #342, opened 2026-01-13
ci: Fix lack of free disk space

sergio-costas:
Currently, CI build fails due to lack of free disk space.

This patch fixes it by removing the Android SDK from the build image.


3v1n0 (approved):
Eheh, good inspection!

PR #343, opened 2026-01-14
fix: The mesa backport was only for amd64, fallback gracefully

kenvandine:
The mesa backport was only for amd64, so fallback gracefully on other architectures.


sergio-costas (approved):
Looks sensible.

PR #344, opened 2026-01-19
Build Runtime and test it

sergio-costas:
This PR builds the runtime against a new SDK, and tests it against several snaps. Specifically, it tests:

* firefox
* thunderbird
* darktable
* cheese
* gimp
* gnome-calculator
* gnome-system-monitor
* gnome-text-editor
* evince
* kicad
* chromium
* shotwell
* telegram
* mattermost

It uses YARF OCR capabilities to detect that the application has been launched correctly. After it detects it, it waits five seconds and test again, to ensure that it is still working (thus to avoid the case where the window does appear but the application crashes immediately).

If a test fails, it uploads an artifact with the YARF logs, to check if it's a test problem or a PR problem.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,14 +9,167 @@ on:
 jobs:
   snap:
     runs-on: ubuntu-latest
+    outputs:
+      runtime: ${{ steps.upload-runtime.outputs.artifact-id }}
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v6
+
     - name: Free disk space
       run: |
         sudo rm -rf /usr/local/lib/android
-    - uses: snapcore/action-build@v1
+
+    - name: Build SDK
+      uses: snapcore/action-build@v1
       id: snapcraft
-    - uses: actions/upload-artifact@v4
+
+    - uses: actions/upload-artifact@v6
       with:
         name: 'snap'
         path: ${{ steps.snapcraft.outputs.snap}}
+
+    - name: Prepare runtime
+      run: |
+        set -eu
+        git clone --depth 1 --branch gnome-46-2404 https://github.com/ubuntu/gnome-sdk gnome-runtime
+        mkdir gnome-runtime/base-gnome-sdk
+        mv ${{ steps.snapcraft.outputs.snap}} gnome-runtime/base-gnome-sdk/
+        cd gnome-runtime
+        ./local-build.py --prepare-only
+
+    - name: Build runtime
+      uses: snapcore/action-build@v1
+      id: snapcraft-runtime
+      with:
+        path: gnome-runtime
+
+    - name: Delete old snap
+      run: |
+        rm -f gnome-runtime/base-gnome-sdk/*.snap
+
+    - uses: actions/upload-artifact@v6
+      id: upload-runtime
+      with:
+        name: 'snap-runtime'
+        path: ${{ steps.snapcraft-runtime.outputs.snap}}
+
+  testing:
+    strategy:
+      fail-fast: false
+      matrix:
+        # define in 'snapname' the name of each snap to test
+        snapname:
+          - cheese
+          - chromium
+          - darktable
+          - evince
+          - epiphany
+          - firefox
+          - gimp
+          - gnome-calculator
+          - gnome-system-monitor
+          - gnome-text-editor
+          - kicad
+          - mattermost-desktop
+          - shotwell
+          - telegram-desktop
+          - thunderbird
+        # here, in waittext, define the text to find for each snap
+        # more variables are:
+        # - commandname: the command to run instead of 'snap run $snapname'
+        # - timeout: the timeout to wait for the text to appear. Default is 20 seconds.
+        # - connections: an space-separated list of plugs that must be connected
+        # - enablex11: if true, will enable Xwayland
+        # - separated: if true, the words will be split and must be found each anywhere in the screen
+        include:
+          - snapname: "cheese"
+            waittext: "Take a Photo"
+          - snapname: "chromium"
+            waittext: "New Tab"
+            enablex11: "true"
+          - snapname: "darktable"
+            waittext: "darktable lighttable darkroom"
+            separated: "true"
+          - snapname: "evince"
+            waittext: "Document Viewer Open"
+          - snapname: "epiphany"
+            waittext: "Set as Default Browser"
+            enablex11: "true"
+          - snapname: "firefox"
+            waittext: "Welcome to Firefox"
+          - snapname: "gimp"
+            waittext: "Welcome to GIMP"
+          - snapname: "gnome-calculator"
+            waittext: "Basic mod"
+            separated: "true"
+          - snapname: "gnome-system-monitor"
+            waittext: "Network Resources Processes Disk Memory Swap CPU"
+            separated: "true"
+          - snapname: "gnome-text-editor"
+            waittext: "New Document"
+          - snapname: "kicad"
+            waittext: "Configure KiCad"
+            enablex11: "true"
+          - snapname: "mattermost-desktop"
+            waittext: "Mattermost"
+            separated: "true"
+            commandname: "snap run mattermost-desktop"
+            enablex11: "true"
+            connections: "login-session-observe"
+          - snapname: "shotwell"
+            waittext: "Shotwell"
+          - snapname: "telegram-desktop"
+            waittext: "Welcome to the official Telegram Desktop app"
+          - snapname: "thunderbird"
+            waittext: "Thunderbird will automatically search"
+    runs-on: ubuntu-latest
+    needs: snap
+    steps:
+    - name: Get data
+      uses: actions/checkout@v6
+    - name: Download artifact
+      id: download-snap
+      uses: actions/download-artifact@v8
+      with:
+        artifact-ids: ${{ needs.snap.outputs.runtime }}
+    - name: Install tools
+      run: |
+        set -eu
+        sudo snap refresh
+        sudo apt update
+        # dbus-x11 is required because it contains dbus-launch!!
+        # x11-apps are required to wait for XWayland
+        sudo apt install -y \
+          dbus-daemon \
+          dconf-gsettings-backend \
+          xwayland \
+          gnome-keyring \
+          wayland-utils \
+          xdg-desktop-portal-gnome \
+          gnome-session-common \
+          gnome-settings-daemon \
+          x11-apps
+        # to be able to use Xwayland, --devmode is required
+        sudo snap install --devmode mir-test-tools
+        sudo snap install --beta yarf
+        sudo snap connect yarf:process-control
+        sudo snap install --dangerous ${{ steps.download-snap.outputs.download-path}}/*.snap
+
+# test application with the new runtime
+    - name: Test application
+      run: |
+        export TEST_SNAPNAME=${{ matrix.snapname }}
+        export TEST_WAITTEXT="${{ matrix.waittext }}"
+        export TEST_COMMAND_NAME="${{ matrix.commandname }}"
+        export TEST_DO_ENABLE_X11=${{ matrix.enablex11 }}
+        export TEST_DO_SEPARATED=${{ matrix.separated }}
+        export TEST_TIMEOUT=${{ matrix.timeout }}
+        export TEST_IFACE_CONNECTIONS=${{ matrix.connections }}
+
+        ./.github/workflows/test_script.sh
+
+    - uses: actions/upload-artifact@v6
+      id: upload-yarf-log
+      if: always()
+      with:
+        path: ~/snap/yarf/common/yarf-outdir
+        name: yarf-logs-${{ matrix.snapname }}
--- a/.github/workflows/test_script.sh
+++ b/.github/workflows/test_script.sh
@@ -1,0 +1,140 @@
+#!/bin/bash
+
+set -eux -o pipefail
+
+export TEST_COMMAND_NAME="${TEST_COMMAND_NAME:-snap run ${TEST_SNAPNAME}}"
+# default value for TEST_TIMEOUT: 20 seconds
+export TEST_TIMEOUT=${TEST_TIMEOUT:-20}
+
+echo Testing ${TEST_SNAPNAME} with $TEST_COMMAND_NAME and waiting for ${TEST_WAITTEXT}
+rm -rf ./testRobot
+mkdir ./testRobot
+
+TEST_ROBOT_FILE=./testRobot/app.robot
+cat > $TEST_ROBOT_FILE <<EOF
+*** Settings ***
+Resource        kvm.resource
+Library         Process
+
+*** Tasks ***
+Do test
+EOF
+if [[ "${TEST_DO_SEPARATED}" == "true" ]]; then
+    for entry in ${TEST_WAITTEXT}; do
+    echo "  Match Text    ${entry}    ${TEST_TIMEOUT}" >> ${TEST_ROBOT_FILE}
+    export TEST_TIMEOUT=""
+    done
+    echo "  Sleep    5" >> ${TEST_ROBOT_FILE}
+    for entry in ${TEST_WAITTEXT}; do
+    echo "  Match Text    ${entry}" >> ${TEST_ROBOT_FILE}
+    done
+else
+    cat >> $TEST_ROBOT_FILE <<EOF
+    Match Text    ${TEST_WAITTEXT}    ${TEST_TIMEOUT}
+    Sleep    5
+    Match Text    ${TEST_WAITTEXT}
+EOF
+fi
+cat >> $TEST_ROBOT_FILE <<EOF
+    Keys Combo    Alt_L    F4
+    [Teardown]    Run Keyword    Log Screenshot
+EOF
+
+echo "Test robot file:"
+cat ${TEST_ROBOT_FILE}
+sudo snap install ${TEST_SNAPNAME} || echo Snap already installed
+# This is needed because the gnome runtime has been installed manually with --dangerous
+sudo snap connect ${TEST_SNAPNAME}:gnome-46-2404 gnome-46-2404:gnome-46-2404
+for plug in ${TEST_IFACE_CONNECTIONS}; do
+    sudo snap connect ${TEST_SNAPNAME}:${plug}
+done
+sudo /usr/lib/snapd/snap-discard-ns ${TEST_SNAPNAME}
+echo Current connections:
+snap connections ${TEST_SNAPNAME}
+echo
+
+MYUID=$(id -u)
+export WAYLAND_DISPLAY=/run/user/${MYUID}/wayland-0
+export MIR_SERVER_ADD_WAYLAND_EXTENSIONS=zwlr_screencopy_manager_v1:zwlr_virtual_pointer_manager_v1
+export DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/${MYUID}/bus
+export LANG=en_US.UTF-8
+export XDG_CONFIG_DIRS=/etc/xdg/xdg-ubuntu:/etc/xdg
+export XDG_MENU_PREFIX=gnome-
+export XDG_SESSION_DESKTOP=ubuntu
+export XDG_SESSION_TYPE=wayland
+export XDG_CURRENT_DESKTOP=ubuntu:GNOME
+export XDG_SESSION_CLASS=user
+export XDG_RUNTIME_DIR=/run/user/${MYUID}
+export XDG_DATA_DIRS=/usr/share/ubuntu:/usr/share/gnome:/usr/local/share/:/usr/share/:/var/lib/snapd/desktop
+export GNOME_SHELL_SESSION_MODE=ubuntu
+export GNOME_SETUP_DISPLAY=:1
+
+# use only GTK portal, since mutter isn't running
+sudo sed -i -e 's/gnome;gtk;/gtk;/g' /usr/share/xdg-desktop-portal/gnome-portals.conf
+
+# this must be defined after launching the demo server; if not, it won't launch
+unset DISPLAY
+export ENABLE_X11_PROP=""
+if [[ "${TEST_DO_ENABLE_X11}" == "true" ]]; then
+    export ENABLE_X11_PROP="--enable-x11=true"
+fi
+
+# required to work in a no-DRM environment
+export XWAYLAND_NO_GLAMOR=1
+mir-test-tools.demo-server --platform-display-libs mir:virtual --virtual-output 1920x1080 ${ENABLE_X11_PROP} &  MIR_PID=$! # starts the demo Mir compositor
+
+counter=0
+until wayland-info > /dev/null || (( ${counter} > 30 )); do
+    echo waiting for mir server...
+    sleep 1
+    counter=$((counter+1))
+done
+
+if [[ "${TEST_DO_ENABLE_X11}" == "true" ]]; then
+    export DISPLAY=:0
+    counter=0
+    until (( ${counter} > 30 )); do
+        timeout 2 xclock || EXIT_CODE=$?
+        # 124 is "timeout"
+        if (( "EXIT_CODE" == "124" )); then
+            # end loop if timed out
+            echo XWayland available
+            counter=50
+        else
+            echo Waiting for XWayland to be available
+            sleep 1
+        fi
+        counter=$((counter+1))
+    done
+    unset WAYLAND_DISPLAY
+    export XDG_SESSION_TYPE=x11
+fi
+
+echo Launching ${TEST_COMMAND_NAME}
+${TEST_COMMAND_NAME} & command_pid=$!
+
+echo Launching YARF
+YARF_RETERR=0
+APP_RETERR=0
+MIR_RETERR=0
+yarf --platform Mir ./testRobot || YARF_RETERR=$?
+
+echo Journal
+journalctl --user --no-pager
+
+echo Waiting for application to die
+# wait up to 10 seconds for the snap to finish.
+timeout 10 tail --pid=${command_pid} -f /dev/null || /bin/true
+
+echo killing MIR server
+# kill DBus and MIR and wait for it to fully finish.
+kill $MIR_PID
+wait $MIR_PID
+
+echo Cleaning up
+# remove user data, just in case
+rm -rf ~/snap/${TEST_SNAPNAME}
+sudo snap remove ${TEST_SNAPNAME} || echo Failed to remove the snap
+rm -rf ./testRobot
+# only fail after cleanup
+exit ${YARF_RETERR}
--- a/README.md
+++ b/README.md
@@ -27,3 +27,21 @@ Core22/Gnome-42-2204:
 Core24/Gnome-46-2404:
 
     git clone -b gnome-46-2404-sdk https://github.com/ubuntu/gnome-sdk.git gnome-46-2404-sdk
+
+## CI tests
+
+The `build.yml` CI test builds the SDK, then builds the RUNTIME from it, and finally
+tests it against a curated list of snaps. It installs the needed dependencies, connects
+the RUNTIME being tested to each snap, launches the snap itself and waits for some specific
+text to appear on the screen. When it happens, it waits five seconds and checks that the
+text is still there, to ensure that the application hasn't crashed.
+
+When adding new tests, the `local-test.py` script can be useful. It uses multipass to
+create a VM and run the test inside. To use it, first initialize the VM with
+
+    local-test.py init
+
+and then you can run the test for a specific snap by calling it with the snap name.
+For example, if you want to test `chromium`, just run
+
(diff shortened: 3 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -13,10 +13,32 @@ jobs:
     - uses: actions/checkout@v2
     - name: Free disk space
       run: |
+        ls /usr/local/share
```
3v1n0: This is just for testing?

.github/workflows/build.yml
```diff
@@ -13,10 +13,32 @@ jobs:
...
       with:
         name: 'snap'
         path: ${{ steps.snapcraft.outputs.snap}}
+    - name: Prepare runtime
+      run: |
+        git clone --depth 1 --branch add-do-not-build-option https://github.com/sergio-costas/gnome-sdk gnome-runtime
+        mv ${{ steps.snapcraft.outputs.snap}} gnome-runtime/gnome-46-2404-sdk.snap
+        cd gnome-runtime
+        ./local-build.py --do-not-build
+        cd ..
```
3v1n0: Is this needed?

.github/workflows/build.yml
```diff
@@ -13,10 +13,32 @@ jobs:
...
+      uses: snapcore/action-build@v1
       id: snapcraft
-    - uses: actions/upload-artifact@v4
+    - uses: actions/upload-artifact@v6
       with:
         name: 'snap'
         path: ${{ steps.snapcraft.outputs.snap}}
+    - name: Prepare runtime
+      run: |
+        git clone --depth 1 --branch add-do-not-build-option https://github.com/sergio-costas/gnome-sdk gnome-runtime
```
3v1n0: This branch will be merged here I guess?

.github/workflows/build.yml
```diff
@@ -13,10 +13,32 @@ jobs:
...
+        mv ${{ steps.snapcraft.outputs.snap}} gnome-runtime/gnome-46-2404-sdk.snap
+        cd gnome-runtime
+        ./local-build.py --do-not-build
```
3v1n0: It seems a bit counter indicative, I'd use `--dry-run` instead

.github/workflows/build.yml
sergio-costas: Ops...yes, it was a test. I'll remove it.

.github/workflows/build.yml
sergio-costas: Hmmm... probably not. I'll check.

.github/workflows/build.yml
sergio-costas: Merged.

.github/workflows/build.yml
sergio-costas: Used your other proposal.

.github/workflows/build.yml
```diff
@@ -14,9 +14,30 @@ jobs:
...
+      uses: snapcore/action-build@v1
       id: snapcraft
-    - uses: actions/upload-artifact@v4
+    - uses: actions/upload-artifact@v6
       with:
         name: 'snap'
         path: ${{ steps.snapcraft.outputs.snap}}
+    - name: Prepare runtime
+      run: |
+        git clone --depth 1 --branch move-base-sdk-snap-into-folder https://github.com/sergio-costas/gnome-sdk gnome-runtime
```
3v1n0: Are we ready to use default branch now?

sergio-costas: As soon as https://github.com/ubuntu/gnome-sdk/pull/350 is merged.

sergio-costas: Done.

3v1n0:
Is this good to go now?

A part squashing the last commit with the initial one, I feel that we're good here.

sergio-costas:
I still want to add the tests themselves: to install several snaps, launch them, and ensure that they run instead of failing.

3v1n0:
> I still want to add the tests themselves: to install several snaps, launch them, and ensure that they run instead of failing.

Fair.. Well, can we just add some very dummy test for now (just ensure that some simple app starts and closes without failing)?

So we can unblock #294 or can we land that without this?

.github/workflows/build.yml
```diff
@@ -11,33 +11,100 @@ jobs:
...
+# install the new runtime and YARF
+    - name: Install runtime
+      run: |
+        sudo snap install yarf
+        sudo snap connect yarf:process-control
+        sudo snap install mir-test-tools
+        sudo snap install --dangerous ${{ steps.snapcraft-runtime.outputs.snap}}
+
+# test applications with the new runtime
+    - name: Test applications
```
3v1n0: Nice, maybe this could be another job though, so we could run it in parallel for each app, and for even more apps in future

3v1n0: Also because if the app test fails (because of a timeout or anything, as yarf can be unreliable), we don't have to perform again the rebuild, but just retry that job.

sergio-costas: How should we do it? Downloading the last artifact?

sergio-costas: Oh, ok... it's simpler...

sergio-costas: Testing using a different job.

.github/workflows/build.yml
```diff
@@ -9,14 +9,114 @@ on:
...
+      run: |
+        rm -f gnome-runtime/base-gnome-sdk/*.snap
+
+    - uses: actions/upload-artifact@v6
+      id: upload-runtime
+      with:
+        name: 'snap-runtime'
+        path: ${{ steps.snapcraft-runtime.outputs.snap}}
+
+  testing:
```
3v1n0, line 55: Yeah, exactly, although this should be split in two jobs (or one taking inputs), so that we repeat for both cheese and gimp and stay as two separated child of the `snap` job

sergio-costas: Ok. First I'll try to ensure that it works :-)

sergio-costas:
@3v1n0 I squashed all the commits, and it's ready for review.

.github/workflows/build.yml
```diff
@@ -7,16 +7,200 @@ on:
...
+  #       cd gnome-runtime
+  #       ./local-build.py --prepare-only
+
+  #   - name: Build runtime
+  #     uses: snapcore/action-build@v1
+  #     id: snapcraft-runtime
+  #     with:
+  #       path: gnome-runtime
+
+  #   - name: Delete old snap
```
3v1n0: I guess we want to revert this test code no?

.github/workflows/build.yml
```diff
@@ -7,16 +7,200 @@ on:
...
+  #     with:
+  #       name: 'snap-runtime'
+  #       path: ${{ steps.snapcraft-runtime.outputs.snap}}
+
+  testing:
+    strategy:
+      fail-fast: false
+      matrix:
+        # define in 'snapname' the name of each snap to test
+        snapname: ["darktable", "cheese", "gimp", "evince", "kicad", "firefox", "thunderbird", "chromium", "shotwell", "telegram-desktop", "mattermost-desktop"]
```
3v1n0: This must be JSON or can be a yaml format?

As I'd prefer to keep it as a multi-line yaml (sorted) list if possible as it's easier to maintain and diff

.github/workflows/build.yml
```diff
@@ -7,16 +7,200 @@ on:
...
+          - snapname: "kicad"
+            waittext: "Configure KiCad"
+            enablex11: "true"
+          - snapname: "evince"
+            waittext: "Document Viewer Open"
+          - snapname: "firefox"
+            waittext: "Welcome to Firefox"
+          - snapname: "thunderbird"
+            waittext: "Thunderbird will automatically search"
+          # - snapname: "epiphany"
```
3v1n0: Maybe add a comment why it's disabled

.github/workflows/build.yml
```diff
@@ -7,16 +7,200 @@ on:
...
     runs-on: ubuntu-latest
+#    needs: snap
     steps:
-    - uses: actions/checkout@v2
-    - name: Free disk space
+#    - name: Download artifact
+#      id: download-snap
+#      uses: actions/download-artifact@v8
+#      with:
+#        artifact-ids: ${{ needs.snap.outputs.runtime }}
```
3v1n0: We should enable these back?

.github/workflows/build.yml
```diff
@@ -7,16 +7,200 @@ on:
...
+    - name: Instal tools
+      run: |
+        sudo snap refresh
+        sudo apt update
+        sudo apt install -y dbus-x11 xwayland gnome-keyring
+        # to be able to use Xwayland, --devmode is required
+        sudo snap install --devmode mir-test-tools
+        sudo snap install --beta yarf
+        sudo snap connect yarf:process-control
+#        sudo snap install --dangerous ${{ steps.download-snap.outputs.download-path}}/*.snap
```
3v1n0: ?

.github/workflows/build.yml
```diff
@@ -7,16 +7,200 @@ on:
...
+        export LANG=en_US.UTF-8
+
+        # this must be defined after launching the demo server; if not, it won't launch
+        unset DISPLAY
+        ENABLE_X11=""
+        if [[ "${{ matrix.enablex11 }}" == "true" ]]; then
+          ENABLE_X11="--enable-x11=true"
+        fi
+        mir-test-tools.demo-server --platform-display-libs mir:virtual --virtual-output 1920x1080 $ENABLE_X11 &  MIRPID=$! # starts the demo Mir compositor
+        sleep 6
```
3v1n0: I'm not in love of sleeps.... Given that yarf will also wait internally, can't we wait for that?

Or can't we just wait until the mir server wayland socket is ready (trying to run `wayland-info` or something until it succeeds, if we don't have better options)?

.github/workflows/build.yml
sergio-costas: Agh, my fault! I was doing tests, and wanted to avoid building over and over the SDK. Let me fix that...

.github/workflows/build.yml
```diff
@@ -7,16 +7,200 @@ on:
...
+        rm -rf ~/snap/${snapname}
+        sudo snap remove ${snapname} || echo Failed to remove the snap
+        rm -rf ./testRobot
+        # only fail after cleanup
+        if [[ "$RETERR" != "0" ]]; then
+          /bin/false
+        fi
+    - uses: actions/upload-artifact@v6
+      id: upload-yarf-log
+      if: failure()
```
3v1n0: These are all the logs, right? Maybe it's worth to keep them all the time though, as it may be useful to check

.github/workflows/build.yml
```diff
@@ -7,16 +7,200 @@ on:
...
+#        sudo snap install --dangerous ${{ steps.download-snap.outputs.download-path}}/*.snap
+
+# test application with the new runtime
+    - name: Test application
       run: |
-        sudo rm -rf /usr/local/lib/android
-    - uses: snapcore/action-build@v1
-      id: snapcraft
-    - uses: actions/upload-artifact@v4
+
```
3v1n0, line 167: `set -eu`

.github/workflows/build.yml
```diff
@@ -7,16 +7,200 @@ on:
...
+        EOF
+        if [[ "${{ matrix.separated }}" == "true" ]]; then
+          for entry in ${waittext}; do
+            echo "  Match Text    ${entry}    ${timeout}" >> ${testrobotfile}
+            unset timeout
+          done
+          echo "  Sleep    5" >> ${testrobotfile}
+          for entry in ${waittext}; do
+            echo "  Match Text    ${entry}" >> ${testrobotfile}
+          done
```
3v1n0: Why not using the match timeout instead of sleeping?

.github/workflows/build.yml
sergio-costas: Fixed. Let me check your other comments...

3v1n0 (commented):
Thanks for this and lots of tests.

There are some things to re-enable, but first I've some cleanups to suggest.

Also, checking the logs, maybe it's also worth to install the xdg-desktop-portal (gtk or gnome)? As they are used by some snaps and in general it's a requirement component to get the snaps working

.github/workflows/build.yml
sergio-costas: Done.

.github/workflows/build.yml
sergio-costas: Yes, it's part of the other piece too.

.github/workflows/build.yml
sergio-costas: It was added for other reason, but the `wayland-info` trick looks promising...

.github/workflows/build.yml
sergio-costas: Yes, are all the logs, including screen captures. If you think that it's better to keep them, I'll do it.

.github/workflows/build.yml
sergio-costas: Done.

.github/workflows/build.yml
sergio-costas: Done.

.github/workflows/build.yml
sergio-costas: Installs the gnome-runtime built from the SDK.

.github/workflows/build.yml
sergio-costas: Done.

.github/workflows/build.yml
```diff
@@ -9,14 +9,213 @@ on:
...
+          - "chromium"
+          - "darktable"
+          - "evince"
+          - "firefox"
+          - "gimp"
+          - "kicad"
+          - "mattermost-desktop"
+          - "shotwell"
+          - "telegram-desktop"
+          - "thunderbird"
```
3v1n0: We should be able to avoid quotes here, but not a big deal.

.github/workflows/build.yml
sergio-costas: I'm already doing that. The idea is to wait five seconds to ensure that the application doesn't crash "a little bit after being launch". So I check the texts twice.

.github/workflows/build.yml
sergio-costas: Done in all parts.

.github/workflows/build.yml
```diff
@@ -9,14 +9,213 @@ on:
...
+        systemctl --user start gnome-keyring-daemon
+        ${commandname} &
+
+        sleep 5
+        RETERR=0
+        yarf --platform Mir ./testRobot || RETERR=$?
+
+        # kill MIR and wait for it to fully finish. It will also kill the snap.
+        kill $MIRPID
+        tail --pid=$MIRPID -f /dev/null
```
3v1n0: Why not just use `wait $MIRPID`

.github/workflows/build.yml
sergio-costas: Removed.

3v1n0:
Maybe temporarily revert https://github.com/ubuntu/gnome-sdk/pull/344/changes/09aca1c8631853445c3b573c1ef0d4dd522393db :)

.github/workflows/build.yml
sergio-costas: Good point!

.github/workflows/build.yml
```diff
@@ -7,16 +7,217 @@ on:
...
+        export snapname=${{ matrix.snapname }}
+        export waittext="${{ matrix.waittext }}"
+        export commandname=${{ matrix.commandname }}
+        export commandname=${commandname:-snap run ${snapname}}
+        echo Testing ${snapname} with $commandname and waiting for ${waittext}
+        rm -rf ./testRobot
+        mkdir ./testRobot
+        # default value for timeout: 20 seconds
+        export timeout=${{ matrix.timeout }}
+        export timeout=${timeout:-20}
```
3v1n0: no need to use `export` here though, they are all local variables.

.github/workflows/build.yml
```diff
@@ -7,16 +7,218 @@ on:
...
+
+        echo "Test robot file:"
+        cat ${testrobotfile}
+        sudo snap install ${snapname} || echo Snap already installed
+        for plug in ${{ matrix.connections }}; do
+          sudo snap connect ${snapname}:${plug}
+        done
+        # This is needed because the gnome runtime has been installed manually with --dangerous
+        sudo snap connect ${snapname}:gnome-46-2404 gnome-46-2404:gnome-46-2404
+        export MYUID=$(id -u ${USERNAME})
```
3v1n0:
```suggestion
        export MYUID=$(id -u)
```

sergio-costas: Done.

.github/workflows/build.yml
sergio-costas: Done.

.github/workflows/build.yml
```diff
@@ -7,16 +7,218 @@ on:
...
+        echo "Test robot file:"
+        cat ${testrobotfile}
+        sudo snap install ${snapname} || echo Snap already installed
+        for plug in ${{ matrix.connections }}; do
+          sudo snap connect ${snapname}:${plug}
+        done
+        # This is needed because the gnome runtime has been installed manually with --dangerous
+        sudo snap connect ${snapname}:gnome-46-2404 gnome-46-2404:gnome-46-2404
+        MYUID=$(id -u)
+        export WAYLAND_DISPLAY=/run/user/${MYUID}/wayland-0
```
3v1n0:
```suggestion
        export WAYLAND_SOCKET=/run/user/${MYUID}/wayland-0
```

3v1n0: O maybe `WAYLAND_DISPLAY=wayland-0`

sergio-costas:
@3v1n0 mattermost was updated this last friday, and now it doesn't launch. It returns a GPU problem. Any idea?

.github/workflows/test_script.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
```
3v1n0, line 2:
```suggestion
set -eux  -o pipefail
```

.github/workflows/test_script.sh
3v1n0, line 1: I would use a clearer name what this is about

.github/workflows/test_script.sh
```diff
@@ -0,0 +1,72 @@
...
+echo Cleaning up
+# remove user data, just in case
+rm -rf ~/snap/${SNAPNAME}
+sudo snap remove ${SNAPNAME} || echo Failed to remove the snap
+rm -rf ./testRobot
+# only fail after cleanup
+if [[ "$YARF_RETERR" != "0" ]]; then
+    /bin/false
+fi
```
3v1n0: nit, add EOF :)

.github/workflows/test_script.sh
```diff
@@ -0,0 +1,72 @@
...
+wait $MIR_PID
+
+echo Cleaning up
+# remove user data, just in case
+rm -rf ~/snap/${SNAPNAME}
+sudo snap remove ${SNAPNAME} || echo Failed to remove the snap
+rm -rf ./testRobot
+# only fail after cleanup
+if [[ "$YARF_RETERR" != "0" ]]; then
+    /bin/false
```
3v1n0: Mh, can't we just `exit 1` here?

.github/workflows/build.yml
```diff
@@ -7,16 +7,213 @@ on:
...
+
+# test application with the new runtime
+    - name: Test application
       run: |
-        sudo rm -rf /usr/local/lib/android
-    - uses: snapcore/action-build@v1
-      id: snapcraft
-    - uses: actions/upload-artifact@v4
+        set -eu
+        set -x
```
3v1n0: let's move this also into another file, at least, as much we can like the robot file stuff

3v1n0:
I would still like to launch a proper dbus session, or even the whole `systemd --user` session where we test, so that we actually have the portals running as we are getting currently:

 ```
 (evince:7698): Gdk-WARNING **: 11:53:00.832: Failed to read portal settings: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: No such interface “org.freedesktop.portal.Settings” on object at path /org/freedesktop/portal/desktop
```

**Comment by sergio-costas** [comment]
> I would still like to launch a proper dbus session, or even the whole `systemd --user` session where we test, so that we actually have the portals running as we are getting currently:

I tried over and over, but it fails due to a cgroups problem.

**Comment by sergio-costas** [comment]
> I would still like to launch a proper dbus session, or even the whole `systemd --user` session where we test, so that we actually have the portals running as we are getting currently

In fact, I'm already configuring the DBUS environment variable to point to the already-running dbus-session instance, and both dconf-daemon and xdg-desktop-portal binaries are being automatically launched, so that error must be due to a different reason.

File: .github/workflows/build.yml (hunk 25 above)
Comment (sergio-costas) [reply .github/workflows/build.yml]: Done.

File: .github/workflows/test_script.sh (hunk 22 above)
Comment (sergio-costas) [reply .github/workflows/test_script.sh:2]: Done.

**Comment by sergio-costas** [comment]
The problem with mattermost is in mattermost itself, so we have to wait for a new version.

**Comment by 3v1n0** [comment]
> > I would still like to launch a proper dbus session, or even the whole `systemd --user` session where we test, so that we actually have the portals running as we are getting currently
> 
> In fact, I'm already configuring the DBUS environment variable to point to the already-running dbus-session instance, and both dconf-daemon and xdg-desktop-portal binaries are being automatically launched, so that error must be due to a different reason.

Could be because the gtk portal does not implement it? Should we try the gnome one?

File: .github/workflows/test_script.sh [hunk 26]
```diff
@@ -0,0 +1,137 @@
...
+    echo waiting for mir server...
+    sleep 1
+    counter=$((counter+1))
+done
+
+if [[ "${TEST_DO_ENABLE_X11}" == "true" ]]; then
+    counter=0
+    set +e
+    until (( ${counter} > 30 )); do
+        timeout 2 xclock
```
3v1n0: You can't still catch the error without dropping `set -e` by using `false || exit_code=$?` syntax.

.github/workflows/test_script.sh
3v1n0: Where are the rest of flags? 😅

.github/workflows/test_script.sh
sergio-costas: I modified it right now by using `set +e` right before, and `set -e` again after. And restored all the flags. Just testing it again... and it works!

sergio-costas:
I restored all the flags.

sergio-costas:
@3v1n0 Maybe the problem with Evince is this? `Apr 15 09:24:34 runnervm35a4x dbus-daemon[14816]: apparmor="DENIED" operation="dbus_method_call"  bus="session" path="/org/freedesktop/portal/desktop" interface="org.freedesktop.host.portal.Registry" member="Register" mask="send" name="org.freedesktop.portal.Desktop" pid=14871 label="snap.evince.evince" peer_pid=15231 peer_label="unconfined"`

sergio-costas:
@3v1n0 Ok, I restored everything. When mattermost is fixed, I think that we can merge this.

.github/workflows/test_script.sh
3v1n0: I know but if possible I would avoid setting /unsettling the flag but rather handle errors everywhere

.github/workflows/test_script.sh
```diff
@@ -0,0 +1,141 @@
...
+kill $MIR_PID
+wait $MIR_PID
+
+echo Cleaning up
+# remove user data, just in case
+rm -rf ~/snap/${TEST_SNAPNAME}
+sudo snap remove ${TEST_SNAPNAME} || echo Failed to remove the snap
+rm -rf ./testRobot
+# only fail after cleanup
+if [[ "$YARF_RETERR" != "0" ]]; then
```
3v1n0:
```suggestion
exit "$YARF_RETERR"
```

sergio-costas: Done.

.github/workflows/test_script.sh
sergio-costas: Not sure that false || exit_code=$? would work as we expect...

.github/workflows/test_script.sh
sergio-costas: What do you propose, instead?

.github/workflows/test_script.sh
sergio-costas: Ok, did a test and yes, it can be done.

.github/workflows/test_script.sh
sergio-costas: Done.

sergio-costas:
@3v1n0 Ok, all tests are passing!!!!! The only open conversation is the name for the YARF logs files.

.github/workflows/test_script.sh
```diff
@@ -0,0 +1,137 @@
...
+export XDG_SESSION_TYPE=wayland
+export XDG_CURRENT_DESKTOP=ubuntu:GNOME
+export XDG_SESSION_CLASS=user
+export XDG_RUNTIME_DIR=/run/user/${MYUID}
+export XDG_DATA_DIRS=/usr/share/ubuntu:/usr/share/gnome:/usr/local/share/:/usr/share/:/var/lib/snapd/desktop
+export GNOME_SHELL_SESSION_MODE=ubuntu
+export GNOME_SETUP_DISPLAY=:1
+
+# use only GTK portal, since mutter isn't running
+sudo sed -i -e 's/gnome;gtk;/gtk;/g' /usr/share/xdg-desktop-portal/gnome-portals.conf
```
aleasto, line 73: mmh, sketchy. we could avoid lying about being a GNOME desktop instead, and it should default to `xdg-desktop-portal-gtk` already.

On the other hand, I know for a fact that some third party applications check for `XDG_CURRENT_DESKTOP=GNOME` before enabling some features so maybe this more realistic than not lying about being GNOME...

sergio-costas: But we aren't cheating, that's exactly the same environment values that Noble returns...

sergio-costas: (it is set to `ubuntu:GNOME`)

aleasto: That's because we're actually running GNOME. Here we're running Mir instead.

sergio-costas: Yes, but I think that we should set an environment as similar as possible to the final one...

aleasto: Yes that's fair

.github/workflows/build.yml
```diff
@@ -9,14 +9,169 @@ on:
...
+        # dbus-x11 is required because it contains dbus-launch!!
+        # x11-apps are required to wait for XWayland
+        sudo apt install -y \
+          dbus-daemon \
+          dconf-gsettings-backend \
+          xwayland \
+          gnome-keyring \
+          wayland-utils \
+          xdg-desktop-portal-gnome \
+          dbus-x11 \
```
3v1n0: Can't we do it only if `enableX11` is set?

3v1n0 (approved):
See that comment if it makes sense, otherwise LGTM.

Also see if some commits (refactoring previous ones) can be squashed.

.github/workflows/build.yml
```diff
@@ -9,14 +9,169 @@ on:
...
+        export TEST_IFACE_CONNECTIONS=${{ matrix.connections }}
+
+        ./.github/workflows/test_script.sh
+
+    - uses: actions/upload-artifact@v6
+      id: upload-yarf-log
+      if: always()
+      with:
+        path: ~/snap/yarf/common/yarf-outdir
+        name: yarf-failure-${{ matrix.snapname }}
```
3v1n0:
```suggestion
        name: yarf-logs-${{ matrix.snapname }}
```

We save it `always()` now.

3v1n0: It looks weird here https://github.com/ubuntu/gnome-sdk/actions/runs/24894424278

.github/workflows/build.yml
sergio-costas: It contains dbus-launch, so that one is required. But x11-apps can be conditional.

.github/workflows/build.yml
sergio-costas: Much better, agree. Done.

.github/workflows/build.yml
aleasto: what uses `dbus-launch`?

sergio-costas:
@aleasto Not sure what uses right now dbus-launch... let's test it.

.github/workflows/build.yml
aleasto: What was the answer?

sergio-costas: The tests that use X11 are failing, so I'm trying again with it put.

sergio-costas: Ok, something happens with MIR and XWayland...

sergio-costas:
OK, I removed dbus-launch, and it works fine, so... It should be ready.

3v1n0 (approved):
Nice thanks a lot for working on this, it will be very relevant and helpful I think (especially when the CI will fail and we have to debug it 😂)!

PR #345, opened 2026-01-20
Use salsa.debian.org for canberra library

sergio-costas:
The canberra library is quite old, so no new updates are expected. For this reason it was downloaded from Lennart Poetering web. Unfortunately, the web seems to have broken, which means that this is fragile.

This patch uses the salsa.debian.org repository instead.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -967,7 +967,9 @@ parts:
 
   libcanberra:
     after: [ libdazzle ]
-    source: http://0pointer.de/lennart/projects/libcanberra/libcanberra-0.30.tar.xz
+    source: https://salsa.debian.org/gnome-team/libcanberra.git
+    source-depth: 1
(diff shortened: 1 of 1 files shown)


nteodosio (commented):
Does that work? I don't see ubuntu/0.30-18ubuntu1 as a branch in https://salsa.debian.org/gnome-team/libcanberra/-/branches/stale.

sergio-costas:
> Does that work? I don't see ubuntu/0.30-18ubuntu1 as a branch in https://salsa.debian.org/gnome-team/libcanberra/-/branches/stale.

Yes, it worked, although as you say, is wrong. I'm changing it to `source-tag`, which is the right one. Let me check again that it works and will push it again...

sergio-costas:
@nteodosio Fixed.

nteodosio (approved):
Thanks!

PR #347, opened 2026-01-27
Add DoNotBuild option to the build script

sergio-costas:
This option just creates the new snapcraft.yaml file but doesn't build the snap. It is required for github CI.

### DIFF
--- a/local-build.py
+++ b/local-build.py
@@ -46,6 +46,8 @@ gnome_part['build-environment'] = gnome_part.get('build-environment', []) + [
 try:
     with open(f'./{MODIFIED_CONFIG}', "w") as config_file:
         config_file.write(yaml.dump(config, Dumper=yaml.Dumper))
+    if "--prepare-only" in sys.argv:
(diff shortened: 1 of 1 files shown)


local-build.py
```diff
@@ -46,6 +46,8 @@ def str_presenter(dumper, data):
 try:
     with open(f'./{MODIFIED_CONFIG}', "w") as config_file:
         config_file.write(yaml.dump(config, Dumper=yaml.Dumper))
+    if "--do-not-build" in sys.argv:
```
3v1n0: Mh maybe rename it in `--prepare-only` or something as it seems weird that we call a local build command negating it afterwards

3v1n0 (approved):
A comment on the naming, otherwise lgtm

local-build.py
sergio-costas: Yes, I like it.

sergio-costas: Done.

PR #348, opened 2026-01-27
Fix clutter source

sergio-costas:
The Clutter repository has been moved to Archive and made read only. This patch fixes the source in the snapcraft.yaml file.


nteodosio (approved):

sergio-costas:
There seems to be even more cases... I'll try to fix all of them and only then merge this.

sergio-costas:
Ok, It is only Clutter and LibDazzle.

PR #349, opened 2026-01-27
Update parts to last versions

sergio-costas:
This include libcroco, because GTK4 requires it and downloads its own version if it's not available in the system, but then librsvg fails to build.

Installing our own libcroco fixes this.

Tested with:

* Firefox
* Cheese
* Chromium
* Darktable
* Element-desktop
* Epiphany
* Evince
* Gimp
* Gnome-Characters
* Gnome-Mines
* Gnome-Recipes
* Kicad
* Shotwell
* Telegram-Desktop
* Mattermost-desktop

All of them work fine.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -74,7 +74,7 @@ parts:
     after: [ ninja ]
     plugin: nil
     source: https://github.com/mesonbuild/meson.git
-    source-tag: '1.10.1'
+    source-tag: '1.10.2'
     source-depth: 1
     override-build: |
       python3 -m pip install --break-system-packages .
@@ -140,7 +140,7 @@ parts:
   glib:
     after: [ libffi, meson-deps ]
     source: https://gitlab.gnome.org/GNOME/glib.git
-    source-tag: '2.86.3'
+    source-tag: '2.86.5'
 # ext:updatesnap
 #   version-format:
 #     ignore-odd-minor: true
@@ -348,7 +348,7 @@ parts:
   gdk-pixbuf:
     after: [ pango, meson-deps, gobject-introspection ]
     source: https://gitlab.gnome.org/GNOME/gdk-pixbuf.git
-    source-tag: '2.44.4'
+    source-tag: '2.44.6'
     source-depth: 1
     plugin: meson
     meson-parameters:
@@ -391,7 +391,7 @@ parts:
   librsvg:
     after: [ libcroco, gdk-pixbuf, vala, meson-deps, gobject-introspection ]
     source: https://gitlab.gnome.org/GNOME/librsvg.git
-    source-tag: '2.61.3' # they left the odd->unstable even->stable scheme, and now all tags are stable
+    source-tag: '2.61.4' # they left the odd->unstable even->stable scheme, and now all tags are stable
 # ext:updatesnap
 #   version-format:
 #     no-9x-revisions: true
@@ -492,7 +492,7 @@ parts:
   libsoup3:
     after: [ libpsl, meson-deps, vala, gobject-introspection ]
     source: https://gitlab.gnome.org/GNOME/libsoup.git
-    source-tag: '3.6.5'
+    source-tag: '3.6.6'
     source-depth: 1
     plugin: meson
     meson-parameters:
@@ -548,7 +548,7 @@ parts:
   gtk3:
     after: [ wayland-protocols, harfbuzz, pango, meson-deps, gobject-introspection ]
     source: https://gitlab.gnome.org/GNOME/gtk.git
-    source-tag: '3.24.51'
+    source-tag: '3.24.52'
 # ext:updatesnap
 #   version-format:
 #     ignore-odd-minor: true
@@ -595,7 +595,7 @@ parts:
   gtk4:
     after: [ libcroco, librsvg, wayland-protocols, harfbuzz, pango, meson-deps, gobject-introspection ]
     source: https://gitlab.gnome.org/GNOME/gtk.git
-    source-tag: '4.20.3'
+    source-tag: '4.20.4'
 # ext:updatesnap
 #   version-format:
 #     ignore-odd-minor: true
@@ -1010,7 +1010,7 @@ parts:
   libinput:
     after: [ libwacom, meson-deps ]
     source: https://gitlab.freedesktop.org/libinput/libinput.git
-    source-tag: '1.30.1'
+    source-tag: '1.30.3'
     source-depth: 1
     plugin: meson
     meson-parameters:
@@ -1122,7 +1122,7 @@ parts:
   libpeas:
     after: [ clutter-gtk, meson-deps, vala, gobject-introspection ]
     source: https://gitlab.gnome.org/GNOME/libpeas.git
-    source-tag: 'libpeas-1.38.0' # developers say that they don't want to break ABI, so it should be safe to always update
+    source-tag: 'libpeas-1.38.1' # developers say that they don't want to break ABI, so it should be safe to always update
 # ext:updatesnap
 #   version-format:
 #     format: 'libpeas-%M.%m.%R'
@@ -1160,7 +1160,10 @@ parts:
   pygobject:
     after: [ pycairo, meson-deps ]
     source: https://gitlab.gnome.org/GNOME/pygobject.git
-    source-tag: '3.55.2'
+# ext:updatesnap
+#   version-format:
+#     ignore-odd-minor: true
+    source-tag: '3.56.2'
     source-depth: 1
     plugin: meson
     meson-parameters:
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -368,10 +369,20 @@ parts:
...
+  libcroco:
+    after: [gdk-pixbuf]
+    source: https://gitlab.com/inkscape/libcroco.git
+    source-depth: 1
+    plugin: autotools
+    autotools-configure-parameters:
+      - --prefix=/usr
+    build-environment: *buildenv
+
+
```
kenvandine: please remove the extra blank line

kenvandine (dismissed):
One inline comment on an extra blank line, otherwise this looks fine.

snapcraft.yaml

nteodosio (approved):

3v1n0:
Still feel it's a bit risky to switch to very new stable versions of gtk and friends, especially because we should also update the themes (yaru) with it.

However, I'd like to get https://github.com/ubuntu/gnome-sdk/pull/344 to land first so at least we can have some CI testing now

sergio-costas:
@3v1n0 If you prefer, I can go back for glib, gobject-introspection, gtk4, glibmm and gtkmm...

snapcraft.yaml
```diff
@@ -347,11 +347,12 @@ parts:
       - -Ddebug=true
       - -Dinstalled_tests=false
       - -Dgtk_doc=false
-      - -Ddocs=false
+      - -Ddocumentation=false
       - -Dintrospection=enabled
       - -Dman=false
       - -Dtests=false
       - -Dinstalled_tests=false
+      - -Dglycin=disabled
```
JGCarroll, line 365: Would it be possible to keep this enabled but add a .patch to make Glycin not use its full sandbox?

That way we get massive boosts to the extensions supported image codecs & gets the memory management boosts from Rust, just lacking some of the extra bubblewrap aspects.

I think it should be possible and I recall that internal assets to the app for example don't use the bubblewrap/flatpak sandbox so you'd basically just nudge the code into treating all files like that.

https://github.com/GNOME/glycin/blob/f96144d7bdf8fb65dface14712795a0b6da6c748/libglycin-rebind/libglycin-rebind/sys/src/lib.rs#L64

Possibly by forcing https://github.com/GNOME/glycin/blob/f96144d7bdf8fb65dface14712795a0b6da6c748/libglycin/src/loader.rs#L47 to always return the NO_SANDBOX enum?

JGCarroll (commented):
Just wondering if we could maybe tweak the Glycin parts on the assumption it might be easy.

snapcraft.yaml
sergio-costas: @aleasto What do you say about this? This is for gdk-pixbuf.

aleasto: Yeah we could provide a patched libglycin as part of the SDK, at least until snapd gives us some options to spawn off an actual new sandbox for glycin. Note that the patched libglycin will work for all C-native projects linking against it, but not for Rust-native projects which instead statically compile against glycin as downloaded from cargo at build time (so they'd still need patching one-by-one). But for the purpose of providing the new gdk-pixbuf, this sounds like a viable option.

I worry though about doing this in the core24 gnome-sdk. While gdk-pixbuf -Dglycin=enabled is ABI compatible, some existing software makes assumptions on the behaviour of the previous loaders so we risk regressions. How about we defer this until a core26 gnome-sdk?

sergio-costas: You are right about the risk of doing it in Core24.

JGCarroll: Given how close that is I'd be happy to wait, we're keen in Pinta for it since it's a huge upgrade to the general codec support even for GTK3 apps; similarly GIMP & etc could stand to gain a lot from it. Although in the aspects of full disclosure we did have an issue in Pinta where the .JPG loader had a subtle change in behaviour that broke things when the Flatpak upgraded to include Glycin. Specifically, saving a JPG with an alpha channel in the original loaders would silently drop the alpha channel as JPG doesn't support it but otherwise work fine, Glycin would throw an explicit error and refuse to do so requiring changes in the app. Most apps would be unlikely to run into that situation but if they did then tying it to a new core has some further merit.

aleasto: > saving a JPG with an alpha channel

Yes, we've seen that exact breakage, as well as some other similar ones, when enabling glycin in Ubuntu 26.04.
Let's wait for core26 so that if we do break applications, they have a chance to notice as part of upgrading the snap base.

sergio-costas:
Ok, these are the stable versions from January 17, so they are the previous stables to the current ones...

snapcraft.yaml
```diff
@@ -130,7 +130,7 @@ parts:
   glib:
     after: [ libffi, meson-deps ]
     source: https://gitlab.gnome.org/GNOME/glib.git
-    source-tag: '2.84.4'
+    source-tag: '2.86.3'
```
aleasto: 2.86.5

snapcraft.yaml
```diff
@@ -65,7 +65,7 @@ parts:
     after: [ ninja ]
     plugin: nil
     source: https://github.com/mesonbuild/meson.git
-    source-tag: '1.8.4'
+    source-tag: '1.10.1'
```
aleasto: 1.10.2

snapcraft.yaml
```diff
@@ -338,7 +338,7 @@ parts:
   gdk-pixbuf:
     after: [ pango, meson-deps, gobject-introspection ]
     source: https://gitlab.gnome.org/GNOME/gdk-pixbuf.git
-    source-tag: '2.42.12'
+    source-tag: '2.44.4'
```
aleasto: 2.44.6, contains CVE fixes

snapcraft.yaml
```diff
@@ -368,10 +369,19 @@ parts:
       - libjpeg-dev
       - libtiff-dev
 
+  libcroco:
+    after: [gdk-pixbuf]
+    source: https://gitlab.com/inkscape/libcroco.git
```
aleasto, line 384: let's pin a commit for reproduceability

snapcraft.yaml
```diff
@@ -528,7 +538,7 @@ parts:
   gtk3:
     after: [ wayland-protocols, harfbuzz, pango, meson-deps, gobject-introspection ]
     source: https://gitlab.gnome.org/GNOME/gtk.git
-    source-tag: '3.24.50'
+    source-tag: '3.24.51'
```
aleasto: 3.24.52

snapcraft.yaml
```diff
@@ -573,9 +583,9 @@ parts:
       patch -p1 < $CRAFT_PROJECT_DIR/patches/gdkglcontext-wayland-Fallback-to-GLES-2_0-after-GL-failed.patch
 
   gtk4:
-    after: [ wayland-protocols, harfbuzz, pango, meson-deps, gobject-introspection ]
+    after: [ libcroco, librsvg, wayland-protocols, harfbuzz, pango, meson-deps, gobject-introspection ]
     source: https://gitlab.gnome.org/GNOME/gtk.git
-    source-tag: '4.18.6'
+    source-tag: '4.20.3'
```
aleasto: 4.20.4

snapcraft.yaml
```diff
@@ -990,7 +1000,7 @@ parts:
   libinput:
     after: [ libwacom, meson-deps ]
     source: https://gitlab.freedesktop.org/libinput/libinput.git
-    source-tag: '1.29.1'
+    source-tag: '1.30.1'
```
aleasto: 1.30.2

snapcraft.yaml
```diff
@@ -1102,7 +1112,7 @@ parts:
   libpeas:
     after: [ clutter-gtk, meson-deps, vala, gobject-introspection ]
     source: https://gitlab.gnome.org/GNOME/libpeas.git
-    source-tag: 'libpeas-1.36.0' # developers say that they don't want to break ABI, so it should be safe to always update
+    source-tag: 'libpeas-1.38.0' # developers say that they don't want to break ABI, so it should be safe to always update
```
aleasto: 1.38.1

snapcraft.yaml
```diff
@@ -1140,7 +1150,7 @@ parts:
   pygobject:
     after: [ pycairo, meson-deps ]
     source: https://gitlab.gnome.org/GNOME/pygobject.git
-    source-tag: '3.54.2'
+    source-tag: '3.55.2'
```
aleasto: 3.56.2

pygobject uses odd version numbers for development releases!

aleasto (changes requested):
I would at least update to the latest patch releases. Some are more important than others.

snapcraft.yaml
sergio-costas: Done.

snapcraft.yaml
sergio-costas: Good point! Added the flag in the part to ensure that snap-updater takes it into account. Thanks!

aleasto (approved):
Thanks.
Let's see what CI thinks 😄

sergio-costas:
CI hates me, for sure :-D

sergio-costas:
@kenvandine What do you say? Approve?

PR #350, opened 2026-01-27
Move base sdk snap into a folder

sergio-costas:
If there are several snaps in the main folder, the build action gets confused and returns the wrong snap (in this case, the SDK snap).

This patch moves the SDK snap into a folder to ensure that only the built runtime snap is in the main folder.

Also ensures not to remove the snapcraft.yaml file when using --prepare-only.

### DIFF
--- a/local-build.py
+++ b/local-build.py
@@ -19,14 +19,25 @@ yaml.representer.SafeRepresenter.add_representer(str, str_presenter) # to use wi
 
 BASE_CONFIG = 'snap/snapcraft.yaml'
 MODIFIED_CONFIG = 'snapcraft.yaml'
-SDK_FILE = 'gnome-46-2404-sdk.snap'
+BASE_SNAP_FOLDER = 'base-gnome-sdk'
+SDK_FILE = None
 
 # If there is a modified snapcraft.yaml, delete it before re-generating it
 if os.path.exists(f'./{MODIFIED_CONFIG}'):
     os.remove(f'./{MODIFIED_CONFIG}')
 
-if not os.path.exists(f'./{SDK_FILE}'):
-    print(f'There is no valid "{SDK_FILE}" file. Aborting.', file=sys.stderr)
+last_time = None
+# get the most recent snap
+for f in os.listdir(BASE_SNAP_FOLDER):
+    if f.endswith(".snap"):
+        fullpath = os.path.join(BASE_SNAP_FOLDER, f)
+        now_time = os.path.getmtime(fullpath)
+        if (last_time is None) or (now_time > last_time):
+            SDK_FILE = fullpath
+            last_time = now_time
+
+if SDK_FILE is None:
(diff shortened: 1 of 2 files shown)


local-build.py
```diff
@@ -19,14 +19,25 @@ def str_presenter(dumper, data):
...
 # If there is a modified snapcraft.yaml, delete it before re-generating it
 if os.path.exists(f'./{MODIFIED_CONFIG}'):
     os.remove(f'./{MODIFIED_CONFIG}')
 
-if not os.path.exists(f'./{SDK_FILE}'):
-    print(f'There is no valid "{SDK_FILE}" file. Aborting.', file=sys.stderr)
+last_time = None
+# get the most recent snap
+for f in os.listdir(BASE_SNAP_FOLDER):
+    if f.endswith(".snap"):
```
3v1n0: I guess you can use `glob.iglob` here no?

local-build.py
```diff
@@ -46,10 +57,9 @@ def str_presenter(dumper, data):
...
-        sys.exit(0)
-
-    os.system('snapcraft clean')
-    os.system('snapcraft pack -v')
+    if "--prepare-only" not in sys.argv:
+        os.system('snapcraft clean')
+        os.system('snapcraft pack -v')
 finally:
-    os.remove(f'./{MODIFIED_CONFIG}')
+    if "--prepare-only" not in sys.argv:
```
3v1n0: Maybe adding a quick `argparse` for this would be fine, but otherwise add a local variable to check instead of repeating the strings-based check

3v1n0 (commented):
Just a nit, fine otherwise

local-build.py
```diff
@@ -19,14 +19,25 @@ def str_presenter(dumper, data):
...
+SDK_FILE = None
 
 # If there is a modified snapcraft.yaml, delete it before re-generating it
 if os.path.exists(f'./{MODIFIED_CONFIG}'):
     os.remove(f'./{MODIFIED_CONFIG}')
 
-if not os.path.exists(f'./{SDK_FILE}'):
-    print(f'There is no valid "{SDK_FILE}" file. Aborting.', file=sys.stderr)
+last_time = None
+# get the most recent snap
```
nteodosio: I suggest to document this.

sergio-costas: I added a better comment, and added some notes in the README.md file.

local-build.py
sergio-costas: Used glob.glob.

local-build.py
sergio-costas: Done the variable. The argparse is overkill for this. If we add more parameters in the future I will add it.

sergio-costas:
@3v1n0 I did the changes. Could you re-review it when you have some spare time, please? #344 depends on this.

3v1n0 (approved):

PR #352, opened 2026-03-04
[2204] Update archived repository URIs

3v1n0:
Redirection does not work well on git clone, so let's just adjust the URIs


sergio-costas (approved):
It's a sensible change.

PR #354, opened 2026-03-20
Fix lack of pciaccess development files

sergio-costas:
When building snaps that depend on libpciaccess, the lack of the soft links to the library makes the build system to link to the static version (like in gnome-boxes), which results in a failure.

This patch fixes it.


3v1n0:
I think it's fine, but we should first merge  #294 that was waiting on #344, so if you can unblock that we can go with them all

nteodosio (approved):

sergio-costas:
@3v1n0 This patch only adds two packages, so it should be safe. #344 will still require a little bit more work, so I'm not sure when I'll be able to complete it. I'll try to have it this week.

PR #359, opened 2026-05-20
core26 sdk

aleasto:
gnome extension update: https://github.com/canonical/snapcraft/pull/6278


snapcraft.yaml
sergio-costas, line 914: Adding `gobject-introspection-bin` fixes the building.

```suggestion
      - iso-codes
      - gobject-introspection-bin
```

sergio-costas (changes requested):

snapcraft.yaml
aleasto: This comes after the `gobject-introspection` part so this shouldn't be necessary.

What's the build failure you're seeing?

sergio-costas: Oh, you are right... well, it can't find g-ir-scanner, so adding the gobject-introspection binary path should be enough...

sergio-costas: That, or this... 
image

I'll do a test.

snapcraft.yaml
```diff
@@ -102,6 +102,9 @@ parts:
       sed -i 's#pkgauxdir="#pkgauxdir="$CRAFT_STAGE#' $LIBTOOLIZE
       sed -i 's#pkgltdldir="#pkgltdldir="$CRAFT_STAGE#' $LIBTOOLIZE
       sed -i 's#aclocaldir="#aclocaldir="$CRAFT_STAGE#' $LIBTOOLIZE
+      for bin in aclocal automake; do
+        ln -s $(find $CRAFT_STAGE/usr/bin -type f -name ${bin}*) $CRAFT_STAGE/usr/bin/$bin
+      done
```
3v1n0: There's no way to call the actual scriptlets with a different root prefix, right?

snapcraft.yaml
```diff
@@ -951,33 +933,8 @@ parts:
       - xkb-data
       - yelp-tools
 
-  cogl:
```
3v1n0, line 957: IMHO this should be the first commit of the branch :)

snapcraft.yaml
sergio-costas: Ok, this is odd... I built it in my system (with a "clean") and it built fine...

sergio-costas: Ok, so the problem seems to be... that g-ir-scanner is trying to do a hard link between /tmp and another folder, and in the building VM /tmp and /root are in different partitions...?

image

snapcraft.yaml
aleasto: Projects generally assume that the unversioned binary is available

snapcraft.yaml
sergio-costas: Ok, I did another PR with your same SDK and now it seems to build, so it looks like a transient.

snapcraft.yaml
aleasto: Done

patches/glycin-snap-no-sandbox.patch
```diff
@@ -0,0 +1,214 @@
...
++    "SNAP_COMMON",
++    "SNAP_DATA",
++    "SNAP_INSTANCE_KEY",
++    "SNAP_INSTANCE_NAME",
++    "SNAP_LIBRARY_PATH",
++    "SNAP_NAME",
++    "SNAP_REAL_HOME",
++    "SNAP_REVISION",
++    "SNAP_USER_COMMON",
++    "SNAP_USER_DATA",
```
3v1n0: Can you also add the ones as per https://github.com/canonical/snapd/pull/16016

patches/glycin-snap-no-sandbox.patch
```diff
@@ -0,0 +1,214 @@
+diff --git a/glycin/src/api_common.rs b/glycin/src/api_common.rs
+index 77285117..4a35f6d6 100644
+--- a/glycin/src/api_common.rs
++++ b/glycin/src/api_common.rs
+@@ -22,6 +22,7 @@ pub enum SandboxMechanism {
+ impl SandboxMechanism {
+     pub async fn detect() -> Self {
+         match RunEnvironment::cached().await {
++            RunEnvironment::Snap => Self::NotSandboxed,
```
3v1n0, line 9: I feel some of this patch can be upstreamed, no?

Can we try to propose it to avoid to maintain in a quick moving project?

Maybe the env variables mapping could be adjusted for snap only, but the rest seems quite fine for upstream too.

aleasto:
I believe this is ready for an edge channel.

3v1n0 (commented):
Oh, I had some comments that I did not send, sorry!

.github/workflows/build.yml
```diff
@@ -53,6 +52,7 @@ jobs:
         path: ${{ steps.snapcraft-runtime.outputs.snap}}
 
   testing:
+    if: false # temporarily disable testing until we have snaps with the new runtime
```
3v1n0:
```suggestion
    # FIXME: temporarily disable testing until we have snaps with the new runtime
    if: false
```

patches/glycin-snap-no-sandbox.patch
aleasto: Once Sergio has a version of this patch that actually implements sandboxing through landlock we can try upstreaming.

I wouldn't bother doing so now.

patches/glycin-snap-no-sandbox.patch
aleasto: I actually considered going the other way and removing everything but the strictly necessary, which should be just `LD_LIBRARY_PATH`.

But there might be some value to setting SNAP_* for external observers.

What do you think?

3v1n0: Yeah, I don't think that any library or program involved should care about snap TBF

snapcraft.yaml
```diff
@@ -65,61 +66,43 @@ parts:
...
     source: https://gitlab.freedesktop.org/fontconfig/fontconfig.git
-    source-tag: 2.15.0
+    source-tag: 2.18.0
     source-depth: 1
-    plugin: autotools
-    autotools-configure-parameters:
+    plugin: meson
+    meson-parameters:
       - --prefix=/usr
       - --libdir=/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR
```
sergio-costas: This isn't required for meson, since it already installs everything there.

sergio-costas (changes requested):

aleasto:
https://launchpad.net/~aleasto/aleasto-craft-remote-build/+snap/snapcraft-gnome-core26-sdk-613cec2f47221e77371bc802f38a7fd1
Successful build on most architectures; riscv64 still pending

PR #361, opened 2026-05-25
Backport at-spi2-atk collection fix

mgorse:
This is needed by orca 50. See
https://gitlab.gnome.org/GNOME/orca/-/commit/e56f8753

### DIFF
--- a/patches/at-spi2-atk-collection-attributes.patch
+++ b/patches/at-spi2-atk-collection-attributes.patch
@@ -1,0 +1,36 @@
+From 7c72d61db98da55fe7db84b79a51b124d487f9ce Mon Sep 17 00:00:00 2001
+From: Mike Gorse <mgorse@suse.com>
+Date: Wed, 19 Jul 2023 12:08:01 -0500
+Subject: [PATCH] collection: Fix match testing for attributes
+
+The predicates for ATSPI_COLLECTION_MATCH_TYPE_NONE and
+ATSPI_COLLECTION_MATCH_TYPE_ALL were both broken.
+
+Fixes #127
+---
+ atk-adaptor/adaptors/collection-adaptor.c |  4 ++--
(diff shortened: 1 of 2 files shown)


sergio-costas:
But is this needed by Orca only, or also by the snapped applications? (because Orca isn't snapped). Also, in which version of at-spi2-atk was this added?

Thanks.

mgorse:
This was fixed in at-spi2-core 2.48.4 and 2.50.0, for GNOME 44/45 (at-spi2-atk is now part of at-spi2-core).

A user reported to me that, when using the brave snap on Ubuntu 26.04, navigating by heading using Orca's hotkeys no longer works[1], although it was working for him on 25.10, and the official .deb vrave also does not exhibit this problem. Orca 50 (shipped in 26.04) has a change in the way it uses AT-SPI's collection interface, so, with this change, it would not work correctly when communicating with applications that link a very old version of libatk-bridge and do not have this fix. When I test with the snap, it responds incorrectly to the Orca query, and, when I look at the snap.yml that brave installs, I see a reference to gnome-42, so I suspect that it is being built against this SDK, so it would need a patched libatk-bridge-2.0.so in order to correctly respond to the orca query. So, yes, it is needed by snapped applications that still use this SDK.

[1] https://mstdn.party/@javm94/116627886334512604

sergio-costas:
Thanks. I think that we should update at-spi2-core in both gnome-42-2204 and gnome-46-2404...

sergio-costas:
Mmm... gnome-46-2404 already has a recent version. So it is needed only in gnome-42-2204.

sergio-costas:
@kenvandine @nteodosio Tested with multipass (works), remmina (works), brave browser (works, tested with youtube), and gnome-boxes (work..ish, because it breaks when trying to connect to the VM; but that already happens with the stable runtime; what worked with the stable runtime continues to work with this PR).

Can you review it when you have some spare time, please? Do you know more snaps to test it?

3v1n0 (approved):

PR #363, opened 2026-05-27
Remove deprecated os.system call

sergio-costas:
This PR replaces os.system() with subprocess.call().

### DIFF
--- a/local-build.py
+++ b/local-build.py
@@ -4,6 +4,7 @@
 
 import sys
 import os
+import subprocess
 import yaml
 import glob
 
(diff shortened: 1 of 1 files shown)


aleasto:
Where did you see that `os.system` is deprecated?

sergio-costas:
Mmm... you are right... It's not "deprecated", but just "recommended" to replace it.

```
The [subprocess](https://docs.python.org/3/library/subprocess.html#module-subprocess) module provides more
powerful facilities for spawning new processes and retrieving their results; using that module is recommended
to using this function.
```

https://docs.python.org/3/library/os.html#os.system

PR #364, opened 2026-05-27
Allow to build runtime from custom sdk

sergio-costas:
This patch adapts the script from gnome-46-2404 that allows to build the Runtime from a custom, prebuilt SDK. It is needed to be able to test any PR in the SDK using CI.

It also fixes the CRAFT_ARCH_TRIPLET warnings.

### DIFF
--- a/local-build.py
+++ b/local-build.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python3
+
+# This script builds the gnome runtime snap using a local SDK snap.
+
+import sys
+import os
+import subprocess
+import yaml
+import glob
+
+def str_presenter(dumper, data):
+    """configures yaml for dumping multiline strings
+    Ref: https://stackoverflow.com/questions/8640959/how-can-i-control-what-scalar-form-pyyaml-uses-for-my-data
+    Ensures that the scripts are dumped in the right multiline format """
+    if len(data.splitlines()) > 1:  # check for multiline string
+        return dumper.represent_scalar('tag:yaml.org,2002:str', data, style='|')
+    return dumper.represent_scalar('tag:yaml.org,2002:str', data)
+
+yaml.add_representer(str, str_presenter)
+yaml.representer.SafeRepresenter.add_representer(str, str_presenter) # to use with safe_dum
+
+BASE_CONFIG = 'snap/snapcraft.yaml'
+MODIFIED_CONFIG = 'snapcraft.yaml'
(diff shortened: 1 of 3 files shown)


local-build.py
```diff
@@ -0,0 +1,65 @@
...
+
+BASE_CONFIG = 'snap/snapcraft.yaml'
+MODIFIED_CONFIG = 'snapcraft.yaml'
+BASE_SNAP_FOLDER = 'base-gnome-sdk'
+SDK_FILE = None
+PREPARE_ONLY_OPTION = '--prepare-only'
+
+# If there is a modified snapcraft.yaml, delete it before re-generating it
+if os.path.exists(f'./{MODIFIED_CONFIG}'):
+    os.remove(f'./{MODIFIED_CONFIG}')
```
3v1n0, line 29: Why heading space and `. /`?

snap/snapcraft.yaml
```diff
@@ -39,12 +39,25 @@ package-repositories:
...
+        sdk_version=$(sed -n  's/^version:\s*\(.*\)/\1/p' \
+          ${CRAFT_PART_INSTALL}/snap.gnome-42-2204-sdk/manifest.yaml)
+      else
+        echo Using local $CRAFT_PROJECT_DIR/${LOCAL_SDK_SNAP}
+        snap install --dangerous $CRAFT_PROJECT_DIR/${LOCAL_SDK_SNAP}
+
+        cp -a /snap/gnome-42-2204-sdk/current/usr $CRAFT_PART_INSTALL/
+        cp -a /snap/gnome-42-2204-sdk/current/etc $CRAFT_PART_INSTALL/
+        cp -a /snap/gnome-42-2204-sdk/current/var $CRAFT_PART_INSTALL/
+        cp -a /snap/gnome-42-2204-sdk/current/lib $CRAFT_PART_INSTALL/
```
3v1n0, line 59: Wouldn't symlink work as well?

aleasto: Might work, but It's better if we keep the same directory structure as the non-local-build.

local-build.py
sergio-costas: Where is a heading space...?

sergio-costas: Do you mean in the comment?

sergio-costas: About the `./`: when the path is passed as relative, I like to put it in the code to emphasize that: `MODIFIED_CONFIG` must be relative to the current folder, never an absolute path.

PR #365, opened 2026-06-01
core26 runtime

aleasto:


aleasto:
I'm going to merge this early so that we can get some more CI running in the SDK merge-request.

PR #366, opened 2026-06-02
Fix ARM64 build

sergio-costas:
In ARM64, a dangling symlink for libdrm_intel remains.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1457,6 +1457,9 @@ parts:
       - libdrm-common
       - libdrm2
       - libdrm-dev
+# needed because libdrm-dev ships a link to libdrm-intel.so, but doesn't depend on it,
+# so we have a broken link.
+      - libdrm-intel1
       - libegl1-mesa-dev
       - libelf1t64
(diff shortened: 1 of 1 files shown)


aleasto:
Let's avoid special-casing libraries; this would be a first as far as I can see...

Instead, I'm starting a local arm64 build to debug why this happens.

aleasto:
> Let's avoid special-casing libraries; this would be a first as far as I can see...
> 
> Instead, I'm starting a local arm64 build to debug why this happens.

The libdrm packaging is broken. `libdrm-dev` ships the `libdrm_intel.so` dev symlink without requiring the `libdrm-intel1` package: https://salsa.debian.org/xorg-team/lib/libdrm/-/merge_requests/11

I suggest explicitly adding `libdrm-intel1` instead as a temporary measure. It's all going to be removed on the runtime side by the mesa cleanup script anyways.

sergio-costas:
@aleasto Done.

aleasto (approved):
Verified locally, LGTM thanks

PR #367, opened 2026-06-02
Fix libwoff 

sergio-costas:
For some reason, libwoff isn't available anymore when building the SDK/Runtime, but it is required for Epiphany (this is why new builds don't pass the CI tests). This patch fix this.

It also removes the bash-completion files, which aren't needed, neither in the SDK nor in the runtime.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1559,6 +1559,7 @@ parts:
       - libwayland-dev
       - libwebkit2gtk-4.1-dev
       - libwebp-dev
+      - libwoff1
       - libwrap0
       - libwrap0-dev
       - libx11-xcb-dev


sergio-costas:
It is adding a lot of files that are already in Core24... I have to check why.

sergio-costas:
@aleasto There are no new/missing files against candidate, so this one seems to fix the problem.

snapcraft.yaml
```diff
@@ -1690,6 +1691,7 @@ parts:
       ln -s vala-0.56 usr/lib/vala-current
       ln -s gettext-0.19.8 usr/share/gettext-current
       ln -s gdk-pixbuf-2.0/2.10.0 usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/gdk-pixbuf-current
+      rm -rf usr/share/bash-completion
```
aleasto: shouldn't this go in the `cleanup` part? for example after `rm -rf usr/share/man`

snapcraft.yaml
```diff
@@ -1559,6 +1559,7 @@ parts:
       - libwayland-dev
       - libwebkit2gtk-4.1-dev
       - libwebp-dev
+      - libwoff1
```
aleasto, line 1562: If we're not going to ship the `-dev` package, I wonder if this should be declared in the runtime snap only.

aleasto (commented):
Please re-commit this with one commit for bash completions and one commit for libwoff

snapcraft.yaml
sergio-costas: That's a good point too. But... maybe we need it for development? Should I include the -dev instead?

sergio-costas: Until I add the package building, we can't be sure...

aleasto: So `libwoff1` used to be a dependency of `libwebkit2gtk-4.1` (which we ship in the SDK and runtime -- side note: maybe we shouldn't?), but it stopped being required in version `2.52.3-0ubuntu0.24.04.1`: this explains why the library suddenly disappeared without ourselves explicitly dropping it.

What's still unclear is how this ends up in epiphany's dependency chain. Do you have the error message?

sergio-costas: Now that I think about it, I don't think it's needed: the header files don't seem to be there.

sergio-costas: Also, removed the bash-completion part. I'll prepare a different PR, as you suggested.

aleasto (approved):
With reasoning in https://github.com/ubuntu/gnome-sdk/pull/367#discussion_r3347282402, makes sense to add it back to the SDK

snapcraft.yaml
sergio-costas: Anyway, the library is currently being distributed in the SDK, so I think that we should preserve that.

aleasto: Yes, absolutely

sergio-costas: This is the error, from the CI test: https://github.com/ubuntu/gnome-sdk/actions/runs/26808560281/job/79046340591?pr=366#step:5:303

```
 libpxbackend-1.0.so: cannot open shared object file: No such file or directory
Failed to load module: /home/runner/snap/yarf/common/.cache/gio-modules/libgiolibproxy.so
libpxbackend-1.0.so: cannot open shared object file: No such file or directory
Failed to load module: /home/runner/snap/epiphany/common/.cache/gio-modules/libgiolibproxy.so
/snap/epiphany/240/usr/bin/epiphany: error while loading shared libraries: libwoff2dec.so.1.0.2: cannot open shared object file: No such file or directory
```

aleasto: Ok, final findings:
libwoff1 was also a dependency of `libwebkitgtk-6.0-4` until version `2.52.3-0ubuntu0.24.04.1`.

`libwebkitgtk-6.0-4`  is not part of the SDK so epiphany stages it itself. But because `libwoff1` was part of the SDK, the `remove_common.py` script run by epiphany would avoid staging it, and hence it breaks when `libwoff1` is not part of the SDK anymore.

PR #368, opened 2026-06-03
Remove bash-completion files

sergio-costas:
The bash-completion files aren't needed, so we can safely remove them and save a little bit of space. Also, when the future test for added and removed files is added, there will be less things that will trigger false alarms.


aleasto (approved):
LGTM, thanks

PR #370, opened 2026-06-08
Add libc6 to fix broken symlink

sergio-costas:
Libtool part stages several packages, including libc6-dev indirectly due to dependencies. This results in a broken link for libc6, which breaks Risc-v and Armhf build.

This PR fixes it. Tested in a Raspberry Pi with Ubuntu server 32 bits.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -87,6 +87,7 @@ parts:
     stage-packages:
       - libtool
       - libtool-bin
+      - libc6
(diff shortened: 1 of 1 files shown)


aleasto (approved):

aleasto:
> which breaks Risc-v and Armhf build.

And causes static linking (which just happens to be broken in riscv and armhf) on any architecture that this is happening on.

PR #371, opened 2026-06-09
Ensure no broken symlinks in stage

sergio-costas:
This PR adds a tool to ensure that there are no broken symlinks in stage, showing which packages must be added if needed.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -41,7 +41,16 @@ parts:
       - CRAFT_EXT_CORE_LEVEL: core24
       - PATH: $CRAFT_STAGE/usr/bin:$PATH
 
+  apt-file:
+    plugin: nil
+    build-packages:
+      - apt-file
+    override-pull: |
+      set -eux
+      apt-file update
+
   ninja:
+    after: [ apt-file ]
     plugin: nil
     source: https://github.com/ninja-build/ninja.git
     source-tag: 'v1.12.1'
@@ -95,6 +104,7 @@ parts:
       sed -i 's#pkgauxdir="#pkgauxdir="$CRAFT_STAGE#' $LIBTOOLIZE
       sed -i 's#pkgltdldir="#pkgltdldir="$CRAFT_STAGE#' $LIBTOOLIZE
       sed -i 's#aclocaldir="#aclocaldir="$CRAFT_STAGE#' $LIBTOOLIZE
+      PYTHONDONTWRITEBYTECODE=1 $CRAFT_PROJECT_DIR/tools/check_broken_links.py $CRAFT_STAGE
 
   fontconfig:
     after: [ libtool ]
@@ -1399,6 +1409,10 @@ parts:
         sed -i 's#libdir=/usr#libdir=${prefix}#' $PC
         sed -i 's#includedir=/usr#includedir=${prefix}#' $PC
       done
+    override-stage: |
+      set -eux
+      craftctl default
+      PYTHONDONTWRITEBYTECODE=1 $CRAFT_PROJECT_DIR/tools/check_broken_links.py $CRAFT_STAGE
 
   debs:
     after:
@@ -1623,6 +1637,10 @@ parts:
       cd $CRAFT_STAGE/usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR
       find . -type f,l -exec rm -fv $CRAFT_PART_INSTALL/usr/lib/{} \;
       find . -type f,l -name "*.so.*" -exec sh -c "rm -fv $CRAFT_PART_INSTALL/usr/lib/{}*" \;
+    override-stage: |
+      set -eux
+      craftctl default
+      PYTHONDONTWRITEBYTECODE=1 $CRAFT_PROJECT_DIR/tools/check_broken_links.py $CRAFT_STAGE
     stage:
       - -usr/bin/fc-cache
 
@@ -1700,24 +1718,11 @@ parts:
       ln -s gettext-0.19.8 usr/share/gettext-current
       ln -s gdk-pixbuf-2.0/2.10.0 usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/gdk-pixbuf-current
 
-      # Check dangling libraries symlinks
-      invalid_links=()
-      core_root_prefix=/snap/$CRAFT_EXT_CORE_LEVEL/current
-      for i in $(find "$CRAFT_PRIME"/{usr/,}lib -xtype l -name '*.so'); do
-        echo "$i: dangling symlink!"
-        invalid_links+=($i)
-      done
-      if [ ${#invalid_links[@]} -gt 0 ]; then
-        echo "Invalid symlinks found:"
-        echo "${invalid_links[@]}"
-        echo "Ensure full package set (runtime and dev files) is staged!"
-        exit 1;
-      fi
-
       # Necessary for armhf builds, triggers review warning
       if [ "$CRAFT_ARCH_TRIPLET_BUILD_FOR" = "arm-linux-gnueabihf" ]; then
         execstack --clear-execstack  usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/libde265.so.0*
       fi
+      PYTHONDONTWRITEBYTECODE=1 $CRAFT_PROJECT_DIR/tools/check_broken_links.py $CRAFT_PRIME
 
   # Used by the gnome snapcraft extensions to load translations from within the content snap
   bindtextdomain:
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -96,6 +96,7 @@ parts:
     stage-packages:
       - libtool
       - libtool-bin
+      - libcrypt1
       - libc6
     override-stage: |
       set -eux
@@ -1643,6 +1644,16 @@ parts:
       PYTHONDONTWRITEBYTECODE=1 $CRAFT_PROJECT_DIR/tools/check_broken_links.py $CRAFT_STAGE
     stage:
       - -usr/bin/fc-cache
+      - to riscv64:
+        - -usr/lib/tcl8.6/graphviz/libtcldot.so
+        - -usr/lib/tcl8.6/graphviz/libtclplan.so
+        - -usr/lib/tcl8.6/graphviz/libtcldot_builtin.so
+        - -usr/lib/tcl8.6/graphviz/libgdtclft.so
+      - to armhf:
+        - -usr/lib/tcl8.6/graphviz/libtcldot.so
+        - -usr/lib/tcl8.6/graphviz/libtclplan.so
+        - -usr/lib/tcl8.6/graphviz/libtcldot_builtin.so
+        - -usr/lib/tcl8.6/graphviz/libgdtclft.so
 
   conditioning:
     after:
--- a/tools/check_broken_links.py
+++ b/tools/check_broken_links.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python3
+
+import glob
+import os
+import sys
+
+def get_package_for_file(filepath):
+    print(f"Searching for package that provides {filepath}")
+    output = os.popen(f"apt-file search -F {filepath}").read()
+    for line in output.splitlines():
+        if ": " in line:
+            package, _ = line.split(": ", 1)
+            if package.find(":") != -1:
+                # if the package name contains ":", it's probably an architecture suffix, so remove it
+                package = package.split(":")[0]
+            return package
+    return None
+
+def main():
+    if len(sys.argv) < 2:
+        print("Usage: check_broken_links.py <search_path>")
+        return 1
+    search_path = sys.argv[1]
+
+    # find broken links for libraries
+    broken_links = []
+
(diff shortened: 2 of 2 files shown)


aleasto:
Uhh I think what you really want here is:
* `apt-file find /path/to/link` to find which package provides the link, and/or
* `apt-file find $(realpath /path/to/link)` to find which package provides the target

3v1n0:
> Uhh I think what you really want here is:
> 
>     * `apt-file find /path/to/link` to find which package provides the link, and/or
> 
>     * `apt-file find $(realpath /path/to/link)` to find which package provides the target

Yeah, but that implies filling a database, can it go using a different root prefix?

aleasto:
> > Uhh I think what you really want here is:
> > ```
> > * `apt-file find /path/to/link` to find which package provides the link, and/or
> > 
> > * `apt-file find $(realpath /path/to/link)` to find which package provides the target
> > ```
> 
> Yeah, but that implies filling a database, can it go using a different root prefix?

If the plan is to run this during the snapcraft build, just call `apt-file update` in override-pull.

tools/check_broken_links.py
```diff
@@ -0,0 +1,216 @@
...
+            else:
+                add_stage_files.append(stage_file)
+
+    # find broken links for libraries
+    broken_links = []
+    ignore_paths = [
+        "usr/lib/gcc",
+        f"usr/lib/{os.environ['CRAFT_ARCH_TRIPLET_BUILD_FOR']}/gcc",
+        "usr/lib/bfd-plugins",
+        f"usr/lib/{os.environ['CRAFT_ARCH_TRIPLET_BUILD_FOR']}/bfd-plugins",
```
aleasto: why?

sergio-costas:
> Uhh I think what you really want here is:
> 
>     * `apt-file find /path/to/link` to find which package provides the link, and/or
> 
>     * `apt-file find $(realpath /path/to/link)` to find which package provides the target

I'm already doing that, but with dpkg...
imagen

Also, using `apt-file` requires to install the package and refresh the cache, something that `dpkg` doesn't require.

aleasto:
> > Uhh I think what you really want here is:
> > ```
> > * `apt-file find /path/to/link` to find which package provides the link, and/or
> > 
> > * `apt-file find $(realpath /path/to/link)` to find which package provides the target
> > ```
> 
> I'm already doing that, but with dpkg... imagen

But that implies that you need to install (and then remove) packages, which we should really avoid.
You also shouldn't assume to have network access at this point of the build.

sergio-costas:
> But that implies that you need to install (and then remove) packages, which we should really avoid. You also shouldn't assume to have network access at this point of the build.

Without network access I can't refresh the package list...

aleasto:
> > But that implies that you need to install (and then remove) packages, which we should really avoid. You also shouldn't assume to have network access at this point of the build.
> 
> Without network access I can't refresh the package list...

You can in `override-pull`

sergio-costas:
> > Without network access I can't refresh the package list...
> 
> You can in `override-pull`

Good point... Let me first debug other parts, and then I'll change that.

snapcraft.yaml
```diff
@@ -1616,6 +1616,11 @@ parts:
         - libdrm-intel1
       - to riscv64:
         - libdrm-intel1
+    override-pull: |
+      set -eux
+      craftctl default
+      apt install apt-file
```
aleasto:
```
build-packages: 
  - apt-file
```

sergio-costas: Yes, I already did. Sorry.

tools/check_broken_links.py
sergio-costas: That was before I took into account the files that were really being staged. Removed.

sergio-costas:
@aleasto It builds armhf, arm64 and amd64. Still waiting for riscv64.

snapcraft.yaml
```diff
@@ -1400,6 +1401,8 @@ parts:
         sed -i 's#includedir=/usr#includedir=${prefix}#' $PC
       done
 
+
+
```
aleasto, line 1417: Nit: spurious whitespace change

snapcraft.yaml
```diff
@@ -1625,6 +1634,20 @@ parts:
...
+        - -usr/lib/tcl8.6/graphviz/libgdtclft.so
+      - to armhf:
+        - -usr/lib/tcl8.6/graphviz/libtcldot.so
+        - -usr/lib/tcl8.6/graphviz/libtclplan.so
+        - -usr/lib/tcl8.6/graphviz/libtcldot_builtin.so
+        - -usr/lib/tcl8.6/graphviz/libgdtclft.so
+    override-prime: |
+      set -eux
+      craftctl default
+      $CRAFT_PROJECT_DIR/tools/check_broken_links.py
```
aleasto: Question: would it make more sense to run this in the conditioning part instead? That already hosts the other dangling links check

aleasto (commented):
Some comments in-line.
Please split the commits into one fixing the build and one adding the assertions

snapcraft.yaml
sergio-costas: Fixed.

snapcraft.yaml
sergio-costas: Good point! Done.

tools/check_broken_links.py
```diff
@@ -0,0 +1,89 @@
...
+
+    # find broken links for libraries
+    broken_links = []
+
+    for filepath in glob.glob(os.path.join(search_path, "**"), recursive=True):
+        if not os.path.islink(filepath):
+            continue
+
+        basename = os.path.basename(filepath)
+        if not (basename.startswith("lib") and ((basename.endswith(".so") or ".so." in basename))):
```
aleasto: I don't think we should assume library names all start with "lib". Did you get false positives without? I would only keep the `.so` matching.

snapcraft.yaml
```diff
@@ -1625,6 +1626,16 @@ parts:
...
+      - to riscv64:
+        - -usr/lib/tcl8.6/graphviz/libtcldot.so
+        - -usr/lib/tcl8.6/graphviz/libtclplan.so
+        - -usr/lib/tcl8.6/graphviz/libtcldot_builtin.so
+        - -usr/lib/tcl8.6/graphviz/libgdtclft.so
+      - to armhf:
+        - -usr/lib/tcl8.6/graphviz/libtcldot.so
+        - -usr/lib/tcl8.6/graphviz/libtclplan.so
+        - -usr/lib/tcl8.6/graphviz/libtcldot_builtin.so
+        - -usr/lib/tcl8.6/graphviz/libgdtclft.so
```
aleasto, line 1656: so I'm curious about this. do you have an explanation?

tools/check_broken_links.py
```diff
@@ -0,0 +1,89 @@
...
+        if ": " in line:
+            package, _ = line.split(": ", 1)
+            if package.find(":") != -1:
+                # if the package name contains ":", it's probably an architecture suffix, so remove it
+                package = package.split(":")[0]
+            return package
+    return None
+
+def main():
+    search_path = os.environ["CRAFT_STAGE"]
```
aleasto: make it a command-line parameter? this could end up useful to run for example in individual parts against `$CRAFT_PART_INSTALL` as well.

snapcraft.yaml
sergio-costas: Basically the package contains them, but there seems to be no package that installs the libraries.

tools/check_broken_links.py
sergio-costas: Good point.

tools/check_broken_links.py
sergio-costas: Mmm... ok, I'll try it.

snapcraft.yaml
sergio-costas: Maybe a leftover...?

aleasto: In `noble` these seem to be packaged in the `libgv-tcl`, which however uses a slightly different path.

So these come from source-built parts? Can you find which part produces them? For example, search `/root/parts` for these files

aleasto: Ah but this is staging in `debs`, so it is not source build. How can it not exist in the archive and yet end up staged in the `debs` part?

tools/check_broken_links.py
sergio-costas: Done.

tools/check_broken_links.py
sergio-costas: Done.

snapcraft.yaml
sergio-costas: I don't know: if I download the -dev packages from armhf and riscv the links are there, but according to apt-cache, no package includes the target files.

snapcraft.yaml
aleasto, line 103: so would it be useful now to run this check?

```suggestion
      $CRAFT_PROJECT_DIR/tools/check_broken_links.py $CRAFT_PART_INSTALL
      craftctl default
```

sergio-costas: It's not needed, because since everything goes into the same stage, it will be catch anyway.

aleasto: True, but we won't know which part was broken (is it `libtool`? is it `debs`?).

It is important to catch this when staging `libtool` because otherwise other parts in this SDK might build against a broken stage, and for example we might never know that we're accidentally statically linking against `libc` because `libtool` staged a broken `libc.so` link.

snapcraft.yaml
aleasto: Ah, so the `libgraphviz-dev` package is simply broken in armhf and riscv64.
Understood, seems like an acceptable workaround.

snapcraft.yaml
aleasto: If we only check the `$CRAFT_PRIME` at the very end of the build, it is still possible that the stage was broken mid-way through the SDK build and it was only fixed by staging some package in the last `debs` part.

sergio-costas: Good point.

sergio-costas: But then, we must check before staging...

sergio-costas: Ok, testing in local...

aleasto: > But then, we must check before staging...

Yes, my suggestion here was to run `$CRAFT_PROJECT_DIR/tools/check_broken_links.py $CRAFT_PART_INSTALL` in `override-stage`, before `craftctl default` (or after, it doesn't actually make a big difference)

sergio-costas: Had same idea. I'm testing in armhf and amd64

sergio-costas: Nope, it has to be in STAGE, and after staging, because doing it in INSTALL will fail because the `stage` part in DEBS is executed after.

aleasto: I don't follow. We want to check `$CRAFT_PART_INSTALL` which is independent of `debs`

aleasto: Alternatively you can also check $CRAFT_STAGE after `craftctl default`

sergio-costas: stage-packages are downloaded into INSTALL, and only copied into STAGE during the stage part, taking into account the `stage:` rules.

Also, in `debs:` part we have the broken links `usr/lib/tcl8.6/graphviz/libtcldot.so` et al, so they will be at INSTALL, but not in STAGE thanks to the `stage:` rules. So the check must be done on STAGE, which is where we can "fix" things, not in INSTALL (unless we use `rm -f $CRAFT_PART_INSTALL/xxxx` in `override-build`... which I suspect that you dislike ;-) )

sergio-costas: Oh! Now I realize: you presumed that I meant checking STAGE before `craftctl default`! No, no... I mean checking STAGE after.

aleasto: we don't necessarily need to have the same checks in every part.
`libtool` doesn't have any `stage:` rules so it wouldn't make a difference.
but this is fine

snapcraft.yaml
```diff
@@ -1700,20 +1725,6 @@ parts:
       ln -s gettext-0.19.8 usr/share/gettext-current
       ln -s gdk-pixbuf-2.0/2.10.0 usr/lib/$CRAFT_ARCH_TRIPLET_BUILD_FOR/gdk-pixbuf-current
 
-      # Check dangling libraries symlinks
```
aleasto, line 1703: I would feel more comfortable with one final check of the CRAFT_PRIME area, here in conditioning

sergio-costas: Ok.

snapcraft.yaml
```diff
@@ -84,17 +84,24 @@ parts:
 # ext:updatesnap
 #   version-format:
 #     ignore: true
+    override-pull: |
+      set -eux
+      apt-file update
```
aleasto: maybe let's add a dedicated part like:

```yaml
apt-file:
  plugin: nil
  build-packages:
    - apt-file
  override-pull: |
    set -eux
    apt-file update
```

sergio-costas: Ok.

aleasto (approved):
awesome, LGTM if CI is green

aleasto (approved):
Ouch, yes

sergio-costas:
Finally!!!! (tested locally. The trick was that environment variable).

PR #376, opened 2026-08-26
[core24] snapcraft/gsettings-desktop-schemas: Add accent colors support

3v1n0:
We used to compile the settings schemas without accent color support, thus snaps might be not aware of it, and in particular of the ubuntu-only brown accent.

Also make libadwaita to depend on this, since it uses the GDesktopAccentColor enum.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -120,7 +120,7 @@ jobs:
           - snapname: "telegram-desktop"
             waittext: "Welcome to the official Telegram Desktop app"
           - snapname: "thunderbird"
-            waittext: "Thunderbird will automatically search"
+            waittext: "Add your email address"
     runs-on: ubuntu-latest
     needs: snap
     steps:
--- a/patches/gsettings-desktop-schemas/001-headers-schemas-Backport-font-rendering-from-47.patch
+++ b/patches/gsettings-desktop-schemas/001-headers-schemas-Backport-font-rendering-from-47.patch
@@ -1,0 +1,45 @@
+From: gnome-sdk <noreply@ubuntu.com>
+Date: Wed, 19 Jun 2024 07:00:00 -0700
+Subject: headers/schemas: Backport font-rendering setting from 47
+
+Backport the GDesktopFontRendering enum and the "font-rendering" schema
+key from the 47 series so that the accent-color backport patches (which
+were authored on top of it) apply as-is on the 46.1 sources.
+
+The enum is intentionally introduced with the original "typdef" typo so
+that the following "headers: Fix typo" patch applies unchanged.
+---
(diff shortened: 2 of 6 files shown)


aleasto (commented):
How does this defect manifest in consumers of the SDK?
Shouldn't apps be using the xdg-desktop-portal to retrieve the accent color anyways?

`font-rendering` is also being backported here with no explaination as to why.

.github/workflows/build.yml
```diff
@@ -84,7 +84,7 @@ jobs:
           - snapname: "cheese"
             waittext: "Take a Photo"
           - snapname: "chromium"
-            waittext: "New Tab"
+            waittext: "Chromium Additional Terms of Service"
```
aleasto: I've asked Nathan to remove this on the Chromium side instead (it's an empty box)

3v1n0:
> How does this defect manifest in consumers of the SDK?

This should not particularly change anything, but now the color is exposed as valid one by libadwaita

> Shouldn't apps be using the xdg-desktop-portal to retrieve the accent color anyways?

Yes, but that goes through libadwaita. Also apps may bypass portals and we do provide them access to settings with no problem :(.

> `font-rendering` is also being backported here with no explaination as to why.

I should have mentioned, but the point was to include missing features that are used by the current libadwaita / gtk versions that we have in the sdk

EDIT: I did it now

.github/workflows/build.yml
3v1n0: mh, ok... I can accept being red on this test then.

3v1n0:
> Yes, but that goes through libadwaita. Also apps may bypass portals and we do provide them access to settings with no problem :(.

Note the hunk in `patches/style-Add-wartybrown-accent-support.patch`, so when we fallback to gsettings (as we do mostly), if this is not defined we just don't use it.

```diff
diff --git a/src/adw-settings.c b/src/adw-settings.c
index df971c3..a116960 100644
--- a/src/adw-settings.c
+++ b/src/adw-settings.c
@@ -192,9 +192,11 @@ init_debug (AdwSettings *self,
       self->accent_color = ADW_ACCENT_COLOR_PURPLE;
     } else if (!g_strcmp0 (env, "slate")) {
       self->accent_color = ADW_ACCENT_COLOR_SLATE;
+    } else if (!g_strcmp0 (env, "brown")) {
+      self->accent_color = ADW_ACCENT_COLOR_BROWN;
     } else {
       g_warning ("Invalid accent color %s (Expected one of: blue, teal, green,"
-                 "yellow, orange, red, pink, purple, slate)", env);
+                 "yellow, orange, red, pink, purple, slate, brown)", env);
     }
   }
 }
```

aleasto:
ugh cause we still allow direct access to the host gsettings.
alright...

aleasto (approved):

